PR #1, opened 2018-03-23
make questions in bold

xmarcoied:
- changing the inline-markup of each questions to strong emphasis(bold)


ojwb:
Great, thanks.

PR #2, opened 2018-03-24
first version proposal of Learning to Rank Clickstream Data Mining

changlinzhang:
This is first version proposal of Learning to Rank Clickstream Data Mining from Changlin Zhang.
Looking forward to your reply. Thanks!

### DIFF
--- a/proposal.rst
+++ b/proposal.rst
@@ -59,27 +59,27 @@
 .. http://teom.wordpress.com/2012/03/01/how-to-write-a-kick-ass-proposal-for-google-summer-of-code/
 
 ======================================
-FILLME WITH THE TITLE OF YOUR PROPOSAL
+Learning to Rank Clickstream Data Mining
 ======================================
 
 About You
 =========
 
- * Name: FILLME
+ * Name: Changlin Zhang
 
- * E-mail address: FILLME
+ * E-mail address: vicky.zhangcl@outlook.com
 
- * IRC nickname(s): FILLME
+ * IRC nickname(s): Vicky
 
- * Any personal websites, blogs, social media, etc: FILLME
+ * Any personal websites, blogs, social media, etc: https://changlinzhang.github.io/
 
- * github URL: FILLME
+ * github URL: https://github.com/changlinzhang
 
  * Biography:
 
 .. Tell us a bit about yourself.
 
-FILLME
+I am a senior student majoring in software engineering at Zhejiang Univeristy, China and I am going to be a graduate student at University of South California in August. I want to join the work of Xapian this summer because  it is most closely related to my previous works and interests. I have experiences in developing two search engines, one recommendation system and two deep learing networks of computer vision. I am skilled in C++, Python and Git. I sincerely hope you could consider my application.
 
 Background Information
 ----------------------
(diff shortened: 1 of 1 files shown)


proposal.rst
```diff
@@ -166,27 +185,35 @@ Project Details
```
ojwb, line 185: In this section we're looking for you to dig into the problem, not just superficially sketch out some high level ideas.  We need convincing that you understand the problem, have found an appropriate way to solve it and that you understand that approach and how to implement it.  And we need to understand what the end result is going to look like to judge if your proposed solution is likely to work for users.

You mention "DCM" and "IAM" papers (but haven't cited them so the reader is left unsure exactly which papers you mean), and pointed to two apparently random pieces of software which use the techniques, but none of that really explains what you're actually planning to do, or how it's going to look when implemented within Xapian.

proposal.rst
```diff
@@ -166,27 +185,35 @@ Project Details
...
 favour of your chosen approach?
 
-FILLME
+DBN, DCM and IAM were come up many years ago. Thus, I may also want to follow investigate some recent works to improve omega results while I havent found suitable ones yet.
 
 Please note any uncertainties or aspects which depend on further research or
 investigation.
 
-FILLME
+DBN, DCM and IAM were come up many years ago. Thus, I may also want to follow investigate some recent works to improve omega results while I havent found suitable ones yet.
```
ojwb, line 210: That's the exact same answer for two questions in a row.  If you don't have an answer, better to just say so that cut and paste the same text.

proposal.rst
```diff
@@ -237,7 +264,11 @@ Project Timeline
 .. any university classes or exams, vacations, etc), make sure you include them
 .. in your project timeline.
 
-FILLME
+April 24, 2018 - May 14, 2018	familiarilize myself with xapian by doing some small changes or demoes
+May 15, 2018 - June 6, 2018		improve DBN with EM and test
+June 7, 2018 - June 30, 2018	implement DCM and IAM and test
+July 1, 2018n - July 21, 2018	finalize letor API with omega and test
+July 21, 2018n - August 6, 2018	explore more recent proposed ideas to improve
```
ojwb: This is much too coarse.  As the comment here says:

> The parts should be AT MOST a week long, or else you won't be able to realistically judge how long they might take.  Even a week is too long really.  Try to break larger tasks down into sub-tasks.

But the real problem is the lack of analysis of the problem in the section above.  Without a concrete plan for how you're going to solve the problem, you can't come up with any sort of realistic timescales for doing it.  If you don't have any sort of realistic timescales for it, then how can either you or we have any well-founded confidence that you can pull this off in the 3 months available?

ojwb (commented):
Thanks for your draft.

I've had a read through and made some comments.

The key thing is really a lack of detail about what you're actually proposing.  It seems you've found some interesting approaches, but you haven't really told us how you're going to make them work in the context of the code this needs to build on top of.

We realise that this planning stage isn't easy, but it is important.

proposal.rst
changlinzhang: I apologized for the hurried version. I made the plan more detailed in new version but I have missed the deadline on website to update it. Thus, I wanna submit it here. https://github.com/changlinzhang/xapian-gsoc-proposal/blob/master/proposal.rst

ojwb:
Thanks for the update.

PR #3, opened 2018-03-27
Math aware search - GSoC proposal

ghost:
Please ignore some other files in this PR.  I keep updating this PR as I revise my proposal.

Thank you.


proposal.rst
```diff
@@ -150,43 +197,134 @@ Motivations
...
 
 **Please note any uncertainties or aspects which depend on further research or
 investigation.**
 
-FILLME
+ We have decided to support latex format for documents and queries. This requires
+ conversion of latex format to mathml format. As this function is in itself can be 
+ a project, we plan to use third party tool like latexmathml. But the exact tool 
+ choice is not finalised yet. This feature will be floater in the work pipeline.
+
```
ojwb: Why is there a need to convert to mathml here at all?

My understanding of MathML is that it's an XML format for representing equations.

We presumably actually want some sort of in-memory parse tree of the equation - what's the reason behind your choice of getting that from LaTeX by converting to MathML and then parsing that, rather than just by parsing the LaTeX?

ghost: This is the approach taken in the paper. I didn't think it the way you pointed out.

But this requires writing two types of parser right? 
What do you think of the time allotted for the preprocessing step? Do you think it's reasonable?

ojwb: It would need a parser for LaTeX and one for MathML to support both, but that's probably easier than writing a LaTeX to MathML convertor and a MathML parser - the problem with writing convertors is that you have to handle any conceptual mismatches between the formats (or lose information, e.g. by throwing away things that don't nicely fit into the target format).

The advantage of converting is you can potentially off-load the work to an existing convertor, though I worry you'll hit the problem of that not converting everything faithfully.  So perhaps it would be better to pick a format to support and worry about other formats later anyway.

But I'm not familiar with MathML and perhaps it's actually just LaTeX recast as XML and so trivial to convert.

I've not had a chance to read the references yet, so not sure about the time for the preprocessing step.  It seems plausible without really knowing what's involved though.

ghost: I guess latex to mathml conversion done because mathml has various tags(variables, constants, operators) that makes representing different token types easy in a tree.

ghost: "The advantage of converting is you can potentially off-load the work to an existing convertor, though I worry you'll hit the problem of that not converting everything faithfully."
Right. I remember reading about this point some where in tangent code.

ojwb:
Thanks for the updates.

PR #4, opened 2018-03-27
Implement Bindings for Go

xmarcoied:
My proposal missing many for sure, but I am willing to learn more about SWIG and Xapian. and going to contribute even outside the gsoc program.

That must be late to review, but I wanted to leave the draft around for any further discussion/comments .

### DIFF
--- a/proposal.rst
+++ b/proposal.rst
@@ -59,27 +59,32 @@
 .. http://teom.wordpress.com/2012/03/01/how-to-write-a-kick-ass-proposal-for-google-summer-of-code/
 
 ======================================
-FILLME WITH THE TITLE OF YOUR PROPOSAL
+Implement Bindings for Go
 ======================================
 
 About You
 =========
 
- * Name: FILLME
+ * Name: Marco Younan
 
- * E-mail address: FILLME
+ * E-mail address: xmarcoied@gmail.com
 
- * IRC nickname(s): FILLME
+ * IRC nickname(s): xmarcoied
 
- * Any personal websites, blogs, social media, etc: FILLME
+ * Any personal websites, blogs, social media, etc: https://xmarcoied.github.io
 
- * github URL: FILLME
+ * github URL: https://github.com/xmarcoied
 
- * Biography:
+ * Biography: 
 
-.. Tell us a bit about yourself.
+    I'm Marko Mohsen from Egypt, a 23 years old guy who study computer engineering at University of Ainshams.
+    a pinaist who can write codes. I am so passionate about the internet as my first passion. from building web applications to making the web and internet a better place, secured and interactive.
+    Through college, I found myself in love with programming and contents/competitions. (I joint many, and got qualified to ACM-ACPC the arab regional contest twice)
 
-FILLME
+    I also joint some local activities in my college, start new intiaitives and so. and have been an active member at `sefrwahed <https://facebook.com/zeroone.asu/>`_ for years, and developed sessions for JS/Go before.
+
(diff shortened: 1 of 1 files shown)


proposal.rst
```diff
@@ -237,7 +275,29 @@ Project Timeline
...
+    * week #3 Support Non-class functions (after this, we should use the non-class functions at xapian in xapian-go)
+    * week #4 improve exceptions/error-handling and writing tests for xapian-go
+**July 13 - August 13**
+    * week #1 writing examples for xapian-go (simpleexpand.go , simpleindex.go , simplematchdecide.go , simplesearch.go)
+    * week #2 writing docs (index.rst) for xapian-go APIs , (docs/bindings/go) , (godoc.org/xapian/xapian-go) & writing more tests for xapian-go-testing
+    documentation will be more specific for golang community at godoc.org for golang-users (gophers), also at docs/bindings/go for xapian users.
+    
+    * week #3 (pencil down): develop example with a real set of data and work with go-tooling
+    
+    * week #4 (offical  pencil down date): Fixing minor bugs, work on merging codes, and publish documentations and xapian-go package
```
ojwb, line 300: I'm not entirely convinced this is the right order of attack.

The functor classes which get subclassed by the user (you mention `MatchDecider`, `ExpandDecider`, `ValueRangeProcessor` (though that's deprecated in favour of `RangeProcessor`) but there are a few others as well) are often a bit harder to wrap because of the calling from Go to C++ to Go, and because of object ownership issues,  They also tend to be more advanced parts of the API.  So I'd probably suggest handling them together (because there's a lot of commonality) and later on (because having the API wrapped enough that you can index and run queries is useful to many users, while having the functors all wrapped but not being able to index isn't very useful to anyone).

I also worry a bit that you've just written down something for each week rather than trying to drill down into what work is needed for each subtask and then thinking about how long that might take you.

I'd also encourage you to think about added Go to the "Getting Started with Xapian" guide as part of this project:

https://getting-started-with-xapian.readthedocs.io/en/latest/

The source for this is written in reStructured text, with custom macros that support pulling in example code in different programming languages, and converting names of classes, methods, constants, etc to match the language the examples are for.  There are versions for several programming languages already (the link above has the Python version), and it takes the user through some examples of indexing and searching data in various different ways.

xmarcoied: I believe I mentioned writing docs for the xapian-go package at the end of block3 , (seems not enough time to write the docs?) . 

Well . maybe I am at wrong for highlighting the work this way and not quite sure about independent work behind classes (so, I am leaving enough time before starting the program to explore more about xapian APIs/Classes ), also the plan was going as supporting as much as I can. and priorities went to Query MatchDecider, RangeProcessors (and ValueRangeProcessor as successor)

ojwb: Ah - I didn't read "writing docs (index.rst) for xapian-go APIs" as meaning the getting started guide, I thought you meant adding a fairly simple `index.rst` in the xapian-bindings tree like those for existing languages (e.g. `xapian-bindings/lua/docs/index.rst`).

ojwb:
Thanks.

ffeng1017:
Hello, I have uploaded my proposal before deadline on my Github and GSoC
dashboard. The link is https://github.com/ffengENG/xapian-gsoc-proposal
I am afraid that my proposal has not been received by you. Thank you!

ojwb:
Don't worry - we have your final proposal OK!

I was just closing the PRs that were open for draft versions as we're past the stage where they are useful.

PR #7, opened 2019-04-07
GSOC 2019 Draft proposal

bazinga65:
I would like to get a feedback on this on how can I improve on it. Is there something else to add? Thank you

### DIFF
--- a/proposal.rst
+++ b/proposal.rst
@@ -129,7 +129,7 @@ Summer of Code project will be the main focus of my time. I have no other engage
 
 My summer vacation starts from April 29, and will end around last week of July. During the vacation, I would be available for work and be online on IRC any time I am awake (usually IST 0800 to 2330). But to organize the schedule, I plan to work during Monday-Friday IST 0830 to 1830 (45-50 hours per week) and catch up on weekends if needed. However, when the classes resume, the working hours would be determined by the schedule of the semester.
 
-**Are you applying for other projects in GSoC 2019?  If so, with which
+**Are you applying for other projects in GSoC this year?  If so, with which
 organisation(s)?**
 
 .. We understand students sometimes want to apply to more than one org and
--- a/proposal.rst
+++ b/proposal.rst
@@ -164,10 +164,11 @@ Project Details
 
 The project is divided into 4 subprojects -
 
-1) Many modern file formats are based around the zip file format with XML contents. So using a zip file reading library instead of the unzip program would is the first target. Currently, Omega uses zlib to read gzip compressed Abiword files. To cover these formats - using libarchive is probably a sensible option. Implementing this library is the first subproject.
-2) Currently the extraction of text from a PDF file is done by running external commands pdfinfo and pdftotext on it and reading their output whereas those two commands use libpoppler to do their work. Hence, using libpoppler and saving running two commands per PDF file would be desirable instead of creating two child processes, and two times any set up and tear down libpoppler doe. Implementing this library is the second subproject.
-3) DjVu is a web-centric format which can display documents and images. DjVuLibre can be used as an extractor for these type of files. Implementing this library is the third subproject.
-4) There are number of other file formats which require external filter programs and can be replaced by available libraries. Although if we could not cover all the libraries available, we could give priority to the file formats which are used more than others such that the overall indexing speed can be reduced by a noticeable amount. This is the fourth subproject.
+1) 
+2) Many modern file formats are based around the zip file format with XML contents. So using a zip file reading library instead of the unzip program would is the first target. Currently, Omega uses zlib to read gzip compressed Abiword files. To cover these formats - using libarchive is probably a sensible option. Implementing this library is the first subproject.
+3) Currently the extraction of text from a PDF file is done by running external commands pdfinfo and pdftotext on it and reading their output whereas those two commands use libpoppler to do their work. Hence, using libpoppler and saving running two commands per PDF file would be desirable instead of creating two child processes, and two times any set up and tear down libpoppler doe. Implementing this library is the second subproject.
+4) DjVu is a web-centric format which can display documents and images. DjVuLibre can be used as an extractor for these type of files. Implementing this library is the third subproject.
+5) There are number of other file formats which require external filter programs and can be replaced by available libraries. Although if we could not cover all the libraries available, we could give priority to the file formats which are used more than others such that the overall indexing speed can be reduced by a noticeable amount. This is the fourth subproject.
 
 The approach for making the libraries work on replacement of external filter programs with shared libraries having same functionalities and hence reducing the time required for indexing is to dynamically load at the runtime. Also, to avoid library bugs from crashing omindex, they can be implemented in a subprocess isolated from the parent process. The output of the child subprocess created using fork() syscall could be piped to the parent process. Another issue is if the library ends up in a memory or CPU eating infinite loop. This can be avoided by using sandboxing to put resource constraints on the process so that the loop will eventually terminate.
 
@@ -261,33 +262,30 @@ Project Timeline
 * Week 2 and 3 from June 3rd to June 16th
 	Implementing the libarchive library for reading zip file formats. Testing and documentation along. (This phase may take a longer time than expected as this is the beginning of coding)
 
-* Week 4 from June 17th to June 23rd
+					 ----- Phase 1 Evaluation -----
+					 
+* Week 5 from June 17th to June 23rd
 	Implementing the libpoppler library for reading zip file formats. Testing and documentation along. Getting libarchive and libpoppler to work before Phase I evaluation will be my main goal.
 
-					 ----- Phase 1 Evaluation -----
-
-* Week 5 and 6 from June 24th to July 7th
+* Week 6 and 7 from June 24th to July 7th
 	Implementing the DjVuLibre library for reading DjVu format. Testing and documentation along.
 
-* Week 7 from July 8th to July 14th
+* Week 8 from July 8th to July 14th
 	Although the libarchive would be able to read the AbiWord files, the extent of AbiWord is large. Hence, I would like to implement the library libabw in case the library crashes. This week can be devoted to it. Testing and documentation all along.
-
-* Week 8 from July 15th-July 21st
+	
+					 ----- Phase 2 Evaluation -----
+					 
+* Week 9 from July 15th-July 21st
 	Implementing other libraries which are available after discussing with mentors. The overall number of libraries to be implemented can be less than the available but I think we could focus on the file formats which are used more than others.
 
-					 ----- Phase 2 Evaluation -----
-
-* Week 9 from July 22nd to July 28th
+* Week 10 from July 22nd to July 28th
 	Continue implementing other libraries.
 
-* Week 10 and 11 from July 29th to August 11th
-	Buffer period: Can be used if any of the previous work is lagged and shifted .Else continue implementing other libraries.
-
-* Week 12 from August 12th - August 18th 
-	Again this period can be used as a buffer period. Final testing and documentation and reviews on the documentation.
+* Week 11 and 12 from August 5th to August 18th
+	Buffer period: Can be used if any of the previous work is lagged and shifted. Else continue implementing other libraries. Start with the final testing documentation.
 
 * Week 13 from August 19th - August 26th
-	Finishing all the remaining tasks (probably testing and documentation) and preparing final report.
+	Finishing all the remaining tasks along with final testing and documentation. Preparing final report.
 
 					 ----- Phase 3 Evaluation -----
 
--- a/proposal.rst
+++ b/proposal.rst
@@ -162,20 +162,14 @@ Project Details
 
 **Describe any existing work and concepts on which your project is based.**
 
-The project is divided into   subprojects - 
-* Implementation of libarchive library for reading formats based around the zip file format instead of running the unzip program.
-* Use of the PDF rendering library libpoppler in the PDF text extractor instead of running external commands pdfinfo and pdftotext.
-* Implementing the DjVuLibre library for reading DjVu format
-* Adding support to other file formats such as AbiWord, Microsoft Publisher, etc. using the libraries available.
-* Using the functionalities of the libraries including loading them dynamically and run in a subprocess to avoid the bugs in the library to be isolated from omindex. 
+The project is divided into 4 subprojects -
 
-1) Many modern file formats are based around the zip file format with XML contents. So using a zip file reading library instead of the unzip program would is the first target. Currently, Omega uses zlib to read gzip compressed Abiword files. To cover these formats - using libarchive is probably a sensible option.
-2) Currently the extraction of text from a PDF file is done by running external commands pdfinfo and pdftotext on it and reading their output whereas those two commands use libpoppler to do their work. Hence, using libpoppler and saving running two commands per PDF file would be desirable instead of creating two child processes, and two times any set up and tear down libpoppler does
-3) DjVu is a web-centric format which can display documents and images. DjVuLibre can be used as an extractor for these type of files.
-4) There are number of other file formats which require external filter programs and can be replaced by available libraries.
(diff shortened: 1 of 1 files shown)


proposal.rst
```diff
@@ -150,14 +146,14 @@ Motivations
...
-FILLME
+This project has many interesting subtasks which I think I will enjoy doing. Also, an open source project is a great way to improve programming skills. There is a lot of scope to think something different and implement it such that to make work. 
 
 **Who will benefit from your project and in what ways?**
 
 .. For example, think about the likely user-base, what they currently have to
 .. do and how your project will improve things for them.
 
-FILLME
+After the implementation of this project, the omindex indexer will use external extractor programs which are available as shared libraries. These libraries will be loaded dynamically during the runtime which will avoid the expense of running other external filters and hence, will speed up indexing as dynamic linking is noticeably faster than static linking. Hence, users of omega will benefit as the search time will be reduced.
```
jaylett: I'm somewhat confused by a couple of the statements here…

> dynamic linking is noticeably faster than static linking

While this is sometimes true, I'm not sure how it's relevant to the difference between either dynamic or runtime linking extractors and forking external extractor programs (which are likely dynamically linked).

> users of omega will benefit as the search time will be reduced

How? Surely this project will only affect index time?

proposal.rst
```diff
@@ -166,27 +162,41 @@ Project Details
...
+4) There are number of other file formats which require external filter programs and can be replaced by available libraries. Although if we could not cover all the libraries available, we could give priority to the file formats which are used more than others such that the overall indexing speed can be reduced by a noticeable amount. This is the fourth subproject.
+
+The approach for making the libraries work on replacement of external filter programs with shared libraries having same functionalities and hence reducing the time required for indexing is to dynamically load at the runtime. Also, to avoid library bugs from crashing omindex, they can be implemented in a subprocess isolated from the parent process. The output of the child subprocess created using fork() syscall could be piped to the parent process. Another issue is if the library ends up in a memory or CPU eating infinite loop. This can be avoided by using sandboxing to put resource constraints on the process so that the loop will eventually terminate.
```
jaylett: It feels to me like there is either an initial subproject, or a dependent step of all of the above, to figure out some of the details of how this would work, and that you should put your current thoughts down here. (It's possible some of this is captured elsewhere, but I think it should be in your proposal.) For instance:

 * I assume you'll load the library needed in the subprocess?
 * Are subprocesses the right model on all operating systems? (Or perhaps on some we won't support this approach at all, at least initially?)
 * are you proposing to dynamically link all these libraries? Or to runtime load them?
 * are there portability concerns? (eg: linking or even compiling the libraries on Windows; efficient transfer of data between the child and parent process on different operating systems)
 * are there automated tests you can write to give confidence in how this system will work? (For instance, you could write a minimal library that can exhibit different failures depending on the file it's run over.)

proposal.rst
```diff
@@ -271,7 +326,7 @@ Use of Existing Code
 **If you already know about existing code you plan to incorporate or libraries
 you plan to use, please give details.**
 
-FILLME
+This project requires the use of existing code i.e. to import various open source file reading libraries such as Poppler, DjVuLibre and other available libraries from the Document Liberation Project.
```
jaylett, line 387: How will omega be affected by the licensing of these libraries? What licenses are they under?

proposal.rst
```diff
@@ -259,7 +313,8 @@ For the avoidance of doubt this includes all contributions to our wiki, mailing
 lists and documentation, including anything you write in your project's wiki
 pages.
 
-FILLME
+Yes, I agree to dual-license all your contributions to Xapian under the GNU
```
jaylett: You probably mean "my contributions" rather than "your contributions" :-)

proposal.rst
```diff
@@ -237,7 +247,51 @@ Project Timeline
...
+* Week 4 from June 17th to June 23rd
+	Implementing the libpoppler library for reading zip file formats. Testing and documentation along. Getting libarchive and libpoppler to work before Phase I evaluation will be my main goal.
+
+					 ----- Phase 1 Evaluation -----
+
+* Week 5 and 6 from June 24th to July 7th
+	Implementing the DjVuLibre library for reading DjVu format. Testing and documentation along.
+
+* Week 7 from July 8th to July 14th
+	Although the libarchive would be able to read the AbiWord files, the extent of AbiWord is large. Hence, I would like to implement the library libabw in case the library crashes. This week can be devoted to it. Testing and documentation all along.
```
jaylett: "in case the library crashes" — in case libarchive crashes you'd use libabw? Or…I'm not sure I understand what you're proposing here.

proposal.rst
```diff
@@ -237,7 +247,51 @@ Project Timeline
...
+
+* Week 8 from July 15th-July 21st
```
jaylett: It feels like everything from here out are as-yet undefined stretch goals. One thing I'd suggest is to put some padding _earlier_ in the project, to ensure that there's time to get early work merged. From experience, the first merge of a GSoC project is often a bit fiddly and may take longer than anticipated, so having time before the first evaluation (alongside other work, but pacing it slower because of the need to address comments &c) is a good idea.

jaylett (commented):
This feels like it's going in the right direction, but it'd be good to see some more detail about how you're going to tackle the idea of using a library rather than an external program. Once you have that nailed, each successive library you bring in (although each will no doubt have details to work out) should be relatively straightforward.

proposal.rst
bazinga65: libarchive handles many types of formats but libabw is specific to AbiWord. Hence, the first thought that came to my mind was that having a separate library for AbiWord documents would be better (I assumed that as libabw is made targetting only AbiWord documents, it would be less likely for it to crash than libarchive) if libarchive crashes for an AbiWord file, we can use libabw.

proposal.rst
bazinga65: Oops! Thank you.

proposal.rst
bazinga65: I had thought about it too but I thought that implementing both libarchive and libpoppler before the phase 1 evaluation would be good so I planned accordingly. Now that you suggest, I shall change it.

proposal.rst
bazinga65: I thought the external filter programs are statically linked, in contrast to what the libraries we would be using because "using these instead would avoid the overhead of running an external filter and so speed up indexing" as given in the idea description. If not, then what factors are affecting on the speeding up of indexing in the case of libraries?

jaylett: The overhead of running an external filter is mostly in executing the new program, which will involve replacing the mapped memory space, loading the new program code, doing any relocation of internal pointers, resolving and runtime linking dynamic libraries, initialising anything that needs it, and probably a bunch of other things I've forgotten — all before it gets to run any code.

By contrast, we want to continue running an existing program, so we're just paying the cost of resolving and runtime linking dynamic libraries, and a small amount of initialisation. (If that's still too high a cost, I guess we could deal with the dynamic libraries in the parent process, and carry them across the `fork()` call.)

Static linking has essentially nothing to do with this; that's a build-time alternative to dynamic linking, and means that the executable will contain all the library functions it needs. Dynamic linking by contrast happens at runtime, typically when you execute a new program (as above), although you can do it under command of the program using calls like `dlopen()`.

It may be worth you reading up on this, because understanding dynamic linking is important to using `dlopen()` effectively.

bazinga65: Now I understood it. Although I know the process of dynamic linking, I know very less about doing it under command using calls like dlopen(). I will read about it and change accordingly. Thank you.

proposal.rst
ojwb: I don't think libarchive can read AbiWord documents - https://github.com/libarchive/libarchive/wiki/LibarchiveFormats certainly doesn't list it as a format, and it isn't an archive or compression format, which is what libarchive's target area is (a `.abw` file is actually an XML document).

But anyway, AbiWord is a format already handled without needing any external filter program (via the XmlParser class).  It's also not a widely used format any more - e.g. https://qa.debian.org/popcon-graph.php?packages=abiword shows the number of reported installs by Debian machines running the optional popcon reporting package - the peak was about 6 years ago and it's tailing off significantly now.  I think most people in the FLOSS world who want a word processor use LibreOffice and OpenDocument files currently.  We still want to support AbiWord so people can index existing documents, but it's not a key format.  So I don't think it's useful to put effort into supporting it via a specific or general third-party library.

More generally, it seems dubious to assume that a more general purpose library would be more crash-prone than a more specific purpose library for a particular file format.  I can see your logic, but really it depends so much on how well maintained and tested each library is - for example, a general-purpose format library with a good automated test suite that's used in a widely used application is likely to more solid than a specific format library which has no automated testsuite and isn't used much.

proposal.rst
```diff
@@ -326,7 +326,7 @@ Use of Existing Code
 **If you already know about existing code you plan to incorporate or libraries
 you plan to use, please give details.**
 
-This project requires the use of existing code i.e. to import various open source file reading libraries such as Poppler, DjVuLibre and other available libraries from the Document Liberation Project.
+This project requires the use of existing code i.e. to import various open source file reading libraries such as Poppler, DjVuLibre and other available libraries from the Document Liberation Project which are free softwares under the GNU General Public License version 2 and later.
```
ojwb: If the `debian/copyright` file from the Debian packaging is accurate:

> Poppler is based on a Xpdf fork (updated at Xpdf 3.02), so the license
> of the poppler core library (libpoppler.so.X) is GPL v2 only.
> All the Poppler contributions are licensed as GPL v2+.

Which says that the poppler core library is GPL v2, not GPL v2 and later (i.e. recipients of the code aren't able to upgrade their licence to a newer GPL version).

So your statement here seems to be inaccurate, but the situation (at least for Poppler - I didn't check the others) is not really a problem for what we want to do.

What we really need to be possible is for people to build omega with these libraries and distribute the resulting binaries - if that's not possible, then distros which ship binary packages (which is most of them) can't package the software.  Currently that's the case for our PHP bindings because the PHP licence is incompatible with the GPL licence, and that's annoying for users as they have to build the bindings themselves.  So what we actually need to check is that the libraries are under licences which are **compatible** with the GPL.

proposal.rst
bazinga65: Thanks, I understood. Then we can devote that time for adding support for other formats. One question. How can I set priority for the formats to be added so that I can cover more widely used formats in this project?

proposal.rst
bazinga65: 1) Document Liberation project: "All participating libraries are released inter alia under the Mozilla Public License Version 2.0. The Document Liberation Project has selected this license to ensure that any new knowledge about a file format will be of benefit to all users"
https://www.gnu.org/licenses/license-list.en.html shows that Mozilla Public License Version 2.0 is compatible with the GPL v2.
2) DjVu : "DjVuLibre-3.5 was developed by Leon Bottou and others as a "Derived Work" of the DjVu Reference Library 3.5. As such, it is also subject to the GNU General Public License version 2."
3) From the above source, the license of the poppler core library is GPL v2.

The licenses of all the software libraries used thus compatible with the GPL. Am I correct?

proposal.rst
bazinga65: Details have been added

proposal.rst
ojwb: It's hard to really know what formats our users (and potential users) are actually caring most about, but some are fairly clearly minority formats and others very mainstream.  E.g. djvu is clearly not widely used, but PDF definitely is.

I'd also tend to favour putting effort in to supporting an open format over a proprietary one when the two have similar popularity.

proposal.rst
ojwb: Yes, it looks like they are all compatible.

proposal.rst
bazinga65: So should I skip DjVu and try to cover the formats from Document Liberation project?

proposal.rst
bazinga65: - Yes, the library will be opened using dlopen() and the functions can be loaded using dlsym() after that. 
- What I found out is that Windows doesn't support fork() and CreateProcess() is the closest thing Windows has to fork(), but Windows requires you to specify an executable to run in that process.
The UNIX process creation is quite different to Windows. Its fork() call basically duplicates the current process almost in total, each in their own address space, and continues running them separately which tells us that Windows doesn't support this approach. 
- We are using dlopen() which runtime loads the libraries
- Portability issues can arise for non *nix systems. I would need to study about that. 
- I can surely write a library and run over files to test it but I don't have the knowledge about how the code in the library extracts it. But a minimal library can be written. I'm not sure how much time it would require (firstly, I would need to read about it and write the code).

jaylett: You should include some notes in your final proposal on where you think the project will go based on what you know about this. It's okay to not have complete information! But note what you do know, and include time to read up & plan what you do not.

proposal.rst
ojwb: Probably the latter are more common than the former.

I'm not sure I'd skip anything though, but rather move it later or last (you could make it explicitly a stretch goal).

The first format (whatever it is) will probably be most work as the framework for these plugins needs creating, and after that I'd expect each additional format would be increasingly quicker.  A new format might reveal some shortcoming in the framework which needs adjusting, but that's going to become less likely as more formats get added.  So you could well find you fly through the last few formats.

bazinga65: Yes, that makes sense. I will shift the DjVu format to the end.

proposal.rst
ojwb: I understand how to use `dlopen()` and `dlsym()`, but I'm not really following how your approach would actually work in practice - perhaps you could sketch out an example of what this looks like using a particular library?

Also could you explain why have you picked this approach over the approach used in the prototype patch linked from the project idea?

ojwb:

proposal.rst
bazinga65: I didn't really think of other approaches before picking up. Perhaps, I picked it because it was suggested in the idea description (it was given that use dlopen() or similar) and I saw (tried to understand) the patch well after it. So at the beginning I thought dlopen() can be used. So I read about dlopen() and I thought this may work and carried that approach forward. Can you tell me what are the problems with this approach?

ojwb: The part about `dlopen()` is talking about any libraries which we use **directly** from `omindex` itself - for example, if we want to use `libarchive` to read ZIP files then it would be quite awkward to do that work in a subprocess so we'd probably really have to use libarchive from the parent process, i.e. omindex itself.  We could make use of libarchive optional, but that would mean someone packaging omega would have to decide whether to enable it or not on behalf of all their users.  They'll probably have to enable it as some users will want it, and then all users have to install libarchive even if they don't use it.

If we use dlopen() and dlsym() for libarchive, that allows a packager to have the best of both worlds - they can enable libarchive support, but make the dependency on libarchive optional, and if it's not install omindex would still work.

But as the project idea says, this may not be worth the effort.  It really depends if the library in question is very likely to be installed anyway.  For example, we just hard depend on zlib because so many things depends on it that it's almost certain to be installed anyway.

ojwb: I don't really see this approach working well for something like libpoppler which we are wanting to wrap in a subprocess.

I think you must be picturing some sort of generic code which takes a library name and some function names and calls dlopen() for the library name and then dlsym() for the function names and then calls those functions to extract text from a document.

But the problem is that the APIs for different extraction libraries vary (hugely in some cases), so it seems this would end up having to create some sort of mini-scripting language in which you'd end up specifying the function call sequence to extract text using libpoppler, etc.  And that's inevitably going to be less flexible that just writing some C or C++ code to call the extraction library's API would be.

Also, while most platforms support something equivalent to `dlopen()` and `dlsym()`, the actual API varies - that means more work to support this for all the platforms omindex currently works on.  And we can't fall back to just linking the library at build time, which we could for libarchive in omindex.

The approach taken in the existing patch is that there's a class framework which takes care of running the communications between the parent and subprocess, and each extraction library is wrapped in a small program which uses the client classes from that framework and links to the extraction library (typically dynamically but it could be static and still work).

This works great for packagers - they can just put each of those small programs into its own package (and only that package needs to depend on the extraction library) and the user can choose which they want to install.

And it'll probably even work with the `CreateProcess()` API you mention since it just requires being able to run a program in a subprocess.

ojwb:

PR #8, opened 2019-04-07
GSoC 2019 proposal: Learning to Rank Clickstream Data Mining 

tson1111:
It is my first version proposal of GSoC 2019 Learning to Rank Clickstream Data Mining. Looking forward to your comments. Thanks a lot!


proposal.rst
```diff
@@ -152,14 +177,31 @@ Motivations
...
+I believe with a better click model and multiple model choices , 
+Xapian will provide better service for users.
 
 **Who will benefit from your project and in what ways?**
 
 .. For example, think about the likely user-base, what they currently have to
 .. do and how your project will improve things for them.
 
-FILLME
+At present, users of Omega will get a document list reranked by the existing
```
jaylett: Is this actually true?

proposal.rst
```diff
@@ -168,27 +210,83 @@ Project Details
 
 **Describe any existing work and concepts on which your project is based.**
 
-FILLME
+My work will be based on the GSoC 2017 Project `Clickstream Mining for 
+Learning to Rank training data <https://trac.xapian.org/wiki/GSoC2017/LetorClickstream>`_.
```
jaylett, line 214: Minor point, but since you are also proposing implementing further algorithms from the original paper and it sounds like some from other papers, it'd be good reference them here.

proposal.rst
```diff
@@ -168,27 +210,83 @@ Project Details
...
+effective to improve document ranking. 
+
+I will first work on an abstract class for all click models and then try to implement 
+each model according to relevant papers. 
+
+
+
+**Phase 2: Documentation**
+
+The second phase of my work will be a detailed documentation of each model I
```
jaylett: This is a good idea, but I'd suggest thinking in terms of different documentation for different audiences. People using the clickstream -> letor system don't need to understand the deep detail of each model; they only really need some guidance on how to choose the best one for their situation. People working on the code (developers of future Xapian functionality in this area) will need a deeper understanding, which is either "internals" documentation, or (and this is often sufficient) documentation in the code itself which references papers, explains how particular algorithms are implemented for Xapian, and so forth.

proposal.rst
```diff
@@ -168,27 +210,83 @@ Project Details
...
+I will first work on an abstract class for all click models and then try to implement 
+each model according to relevant papers. 
+
+
+
+**Phase 2: Documentation**
+
+The second phase of my work will be a detailed documentation of each model I 
+implemented. It will provide the Omega users with a brief view of click models and
+pros & cons of each model. I will continue to work on the clickmodel.rst in the
```
jaylett: This is user documentation, so the first kind I was talking about in a previous comment.

proposal.rst
```diff
@@ -168,27 +210,83 @@ Project Details
...
+The second phase of my work will be a detailed documentation of each model I 
+implemented. It will provide the Omega users with a brief view of click models and
+pros & cons of each model. I will continue to work on the clickmodel.rst in the 
+Omega documentation files.
+
+**Phase 3: Evaluation**
+
+The third phase is about the evaluation. With more and more models applied, I 
+believe it will be necessary to provide an evaluation method for the users. I'm 
+going to use CTR estimation for the evaluation of the models. For the method
```
jaylett: You'll need to unpack what this is — is there a paper that you'll be relying on? How are you adapting any existing methodologies to achieve what you need here? What sort of tools are you going to provide users to run an evaluation?

proposal.rst
```diff
@@ -168,27 +210,83 @@ Project Details
...
+The **UBM model**: the user browsing model is from paper `A user browsing model to predict search engine click data from past observations. SIGIR (2008).  <https://dl.acm.org/citation.cfm?id=1390392>`_.
+
 
 **What other approaches to have your considered, and why did you reject those in
 favour of your chosen approach?**
 
-FILLME
+The Federated Click Model (FCM) and Vertical Click Model (VCM)are not appropiate 
+for Xapian because of the SERP (search engine return pages) of Xapian do not 
+provide with vertical contents.
```
jaylett: I had to look up a paper on click models to figure out what this means. (I assumed "vertical" was to do with layout, rather than to do with category or type clustering, which was confusing.) A little bit more discussion of this would make things clearer. (It also opens the question of whether being able to run the query multiple times in OmegaScript — which would enable searches within verticals, but I believe is not currently possible — would be a good idea for the future.)

proposal.rst
```diff
@@ -168,27 +210,83 @@ Project Details
...
+
+I have also considered about the end to end model.However, I can not make sure to
+develop a clear and fairly smooth process all the way through from logging clicks 
+to turning the logs into relevance judgements to training the model to getting 
+queries reranked. So I didn't continue to work on this idea.
```
jaylett: What would such a model involve? An end-to-end _use case_ of Omega is an important part of the project.

proposal.rst
```diff
@@ -239,7 +337,99 @@ Project Timeline
...
+
+**Coding Week 10: July 30–August 5**
+
+* Implementing the CTR evaluation method for click models.
+* Discuss the frequency of evaluation and the threshold of accuray to change the model.
+
+**Coding Week 11: August 5–August 11**
+
+* Documenting how to use the evaluation method for click models.
+* Decide the default click model and implement the machanism of changing the click models.
```
jaylett: I'd recommend implementing a way of selecting between click models when you implement the second click model (otherwise you end up with a bunch of code that cannot be used easily).

proposal.rst
```diff
@@ -239,7 +337,99 @@ Project Timeline
...
+* Adapt the exsiting SimplifiedDBN class to the abstract class.
+* Start working on the DBN class with EM algorithm.
+
+**Coding Week 3: June 11–June 17**
+
+* Final Week in university.
+* Finish the training method for the DBN class.
+* Discuss the DCM model template with mentors.
+* Start writing automated test for the click models.
+* **Deliverable:** DBN model with EM alogorithm
```
jaylett: Particularly for the first deliverable, you should build in some time to get it merged. Once you've got the first chunk merged, subsequent chunks are generally easier — and from experience, it can take a bit of back and forth on the first one. (While small contributions to Xapian ahead of time help, they don't iron out all the bumps in the process.)

I suggest you assume it'll take a couple of weeks, but can happen alongside other work.

jaylett (commented):
This feels like it's going in the right direction. I think you've missed one of the key aspects of the project, which is to get a full example working from generating clickstream data through to re-ranking in Omega. That should be done first. Other than that, have a look at my comments and think about the whole proposal and project (rather than just addressing each one at the point I've commented).

proposal.rst
ojwb: The people who don't need a deep understanding of the system actually split into two subgroups - there's people using omega -> clickstream -> letor, and developers wanting to integrate the clickstream -> letor part by feeding it click logs from their own Xapian-based search (for example, a non-web based search front-end) but not wanting to try to modify any Xapian code.

One of the original ideas behind adding the clickstream logging to omega was that as well as being useful in itself, it would provide a working example of how to implement this, and the analysis part would be usable for click logs generated by things apart from Omega.

proposal.rst
ojwb: There's collapsing which I guess can be used for a simple form of this, but certainly running two searches in one invocation is not currently possible in OmegaScript (one could probably implement this in JavaScript by combining two sets of search results fetched from separate Omega invocations via XMLHttpRequest for example).

However, it's not that **Xapian** SERPs don't provide vertical content, only that **Omega** ones don't.  It's entirely possible to implement a search system using Xapian that does, and people could generate click logs from such a system to feed into this.

I think it's probably better to focus on getting things working well for the common simple case though, especially as we couldn't easily test such models with Omega currently.

proposal.rst
tson1111: Okay, I'll put the reference here

proposal.rst
tson1111: I will think through to develop a method to satisfy different groups of users.

proposal.rst
tson1111: I will add details to it.

proposal.rst
tson1111: Sorry for my misunderstanding. I will change my words and find a clear way to talk about it.

proposal.rst
tson1111: That's true, I'll consider about the details of the mechanism.

proposal.rst
tson1111: Okay. Besides this, I will try to create a PR for Xapian before the GSoC coding begins.

tson1111:
> This feels like it's going in the right direction. I think you've missed one of the key aspects of the project, which is to get a full example working from generating clickstream data through to re-ranking in Omega. That should be done first. Other than that, have a look at my comments and think about the whole proposal and project (rather than just addressing each one at the point I've commented).

Hi Jaylett, I'm afraid I do not fully understand this comment. Did this mean that I need to show the users how to go through the whole process in documentation by giving them a full example?

tson1111:
I have modified my proposal a little bit. I'd appreciate it if you can find some time to review it. Thanks a lot. @jaylett @ojwb

jaylett:
@tson1111 I don't think it's possible to give a full example, because Omega does not support Letor itself. In your proposal you say:

> the re-ranking command of Omega

but what is this command? `xapian-rank` is the only re-ranking mentioned in Vivek's project write-up — but this isn't part of Omega. We need to enable Omega users to have their search results re-ranked using the trained model.

proposal.rst
ojwb: > Besides this, I will try to create a PR for Xapian before the GSoC coding begins.

We really want to see students work through making a change, submitting a PR, and working through the review process before selection - if we see they can manage that, we can have more confidence about progress during the project (which typically involves quite a lot of patch review).  So to make a difference that needs to be in the next 2 weeks, though sooner is better.

ojwb:

ojwb:

PR #9, opened 2019-04-07
Alexander Slesarev's proposal draft.

slesarev-hub:


proposal.rst
```diff
@@ -152,14 +150,14 @@ Motivations
...
-FILLME
+I have read several articles about how Google search works and it is get interesting to me to have some experience in this area.  
 
 **Who will benefit from your project and in what ways?**
 
 .. For example, think about the likely user-base, what they currently have to
 .. do and how your project will improve things for them.
 
-FILLME
+Firstly, I want to get skills of interacting with people inside project. Secondly, it is very important to try to understand code wich written by other people.
```
ojwb: The question is more aiming to start you thinking about other people your work might benefit.

Once we have some idea who we might be writing the code for, we can use that to help make decisions about the code.  For example, if you're writing a word processor, thinking about the expected users can help you decide how important it is that the user interface should be translatable into other languages, the relative importance of various features (e.g. people writing for publications find a "word count" feature important, but for many people it's never used), and so on.

slesarev-hub:
Sorry for this misunderstanding, I've fix it.

пн, 8 апр. 2019 г. в 02:58, Olly Betts <notifications@github.com>:

proposal.rst
```diff
@@ -273,7 +317,7 @@ Use of Existing Code
 **If you already know about existing code you plan to incorporate or libraries
 you plan to use, please give details.**
 
-FILLME
+I do not exactly know the existing way of transferring data between Letor and Omega, I think it could be fit to do it by Python Pandas library. As I have seen, click models released in C++, so I will do it in C++ too because of speed benefits.
```
jaylett, line 338: I'm not convinced you understand the current state of Letor in Xapian and Omega, or what this project would involve. Right now, Omega can record clickstream logs, and there's an analyser that generates training data for Letor. You can then use Xapian's Letor framework to train a model (using one of the rankers we support). One of the things that's needed is to have Omega use Letor to re-rank results.

jaylett (commented):
We expect to see more detail in the timeline, but I think where you should start is in defining the project — which will then inform the tasks you put into the timeline, and help flesh that out.

proposal.rst
```diff
@@ -152,14 +150,17 @@ Motivations
...
 **Who will benefit from your project and in what ways?**
 
 .. For example, think about the likely user-base, what they currently have to
 .. do and how your project will improve things for them.
 
-FILLME
+This project aimed at search engines developers, it could provide kit of ranking algorithms with ability to change dataset automatically.
+I do not have full understandind Xapian structure now, but I think it could be two ways of user interaction with my part of project:
+	1) Letor and Omega as a parts of Xapian will provide set of user functions to Xapian API. So user will have ability to collect data by Omega and rank dataset by one of Letor algorithms.
+	2) Create separate application for managing collecting and ranking process.
```
ojwb, line 163: While you don't need to fully understand Xapian's structure in every detail (not even to complete the project as there are parts of Xapian you just don't need to know about), you really do need to have at least a high level understanding of the structure of the components that your project is directly involved with in order to propose a meaningful project plan.

For example, Omega is a set of programs (under xapian-applications/omega in the source tree) which make use of the Xapian C++ API (under xapian-core) - there isn't really any way Omega can add functions to that API.

ojwb:

PR #10, opened 2019-04-07
Initial proposal

niweshgupta:


proposal.rst
```diff
@@ -1,5 +1,5 @@
 .. This document is written in reStructuredText, a simple and unobstrusive
-.. markup language.  For an introductiont to reStructuredText see:
+.. markup language.  For an introduction to reStructuredText see:
```
ojwb, line 2: Well spotted - I've push a fix to the template for this.

proposal.rst
```diff
@@ -168,27 +192,28 @@ Project Details
...
 
-FILLME
+Xapian supports almost all of the languages other than PyPy. As Xapian has Python bindings, I'll also be able to write the same for PyPy.
+I tried using -builtin function and written few testcases to test the difference in speed and memory usage with and without -builtin. By modifying the typemaps for exception handling, we'll be able to use -builtin for Xapian.
 
 **What other approaches to have your considered, and why did you reject those in
 favour of your chosen approach?**
 
-FILLME
+The only part in my project which can have different approaches is for the trac issue #263. There are two approaches which Olly has suggested Either: doxy2swig.py should combine the comments for all the possible overloaded forms, or we should provide manually written overrides for all cases where these forms exist. I'll go with the former one as it will be much easier to combine the comments and show them.
```
ojwb: There's actually another potential option for this now - SWIG has fairly recently gained the ability to pick up doxygen-format documentation comments from the headers it is parsing and output them in the generated Python wrappers.

It's probably worth trying out this feature to see how it compares to `doxy2swig.py`.

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+	- Modify doxy2swig.py to combine the errors for all possible overloaded forms
+	- Discuss and test whether the error gets resolved by combining comments only
+	- Testing overall bindings after modification
+	- Merge the changed file
+
+
+Week 4: June 17 - June 23 and Week 5: June 24 - June 30
+
+
+	- Create module 'Xapian' to import from PyPy
```
ojwb: So what does this actually involve?

I don't think there's a `-pypy` option for SWIG, so what will this module look like?

Is it created manually or by SWIG?

Once it's working, what is involved in updating it when a new class or method is added to the C++ API?

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+	- Create Makefiles for PyPy to get easily installed with other language bindings
+	- Write testsuite
+
+
+Week 6: July 1 - July 7
+
+
+	- Figure out and write testcases for smoketest
+	- Create smoketest file
```
ojwb, line 345: How would the smoketest for pypy be different to that for python?

niweshgupta: We can start with the smoketest written for python, then we can add more testcases if needed.

proposal.rst
niweshgupta: It needs a bit of research that's why I've kept two weeks for this. Firstly I'll try to extend swig to add support for PyPy. If it turns out to be too tough or time consuming, then manual creation can be preferred. 

I'll have to write wrappers for the C++ API part.

proposal.rst
niweshgupta: Okay, sure.

proposal.rst
niweshgupta: :)

proposal.rst
jaylett: What are the key challenges to wrapping a C++ library for pypy? You should be able to get an idea of this fairly quickly, and at least indicate what sort of work you'll need to do in order to tackle this.

proposal.rst
jaylett: What other testcases do you anticipate being needed? What about pypy is different to cpython that would cause more edge cases or regressions to exist?

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+
+
+Week 7: July 8 - July 14
+
+
+	- Documentation for Pypy bindings
```
jaylett: What documentation do you feel is necessary? While I'm a big fan of having good documentation, there is some documentation for python3 bindings already which should form a good basis for this.

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+	- Gather metadata and description to pass inside setup() in setup.py
+
+
+Week 10: July 29 - August 4
+
+
+	- Create readme file
+	- Discuss and create licence.md file
+	- Generate distribution archive
+	- Upload the generation archive
```
jaylett, line 373: This shouldn't take a week to do what is basically a short discussion (Xapian's current license is GPL2) and pulling together some documentation. The last two steps are literally two commands: building the wheel and pushing it to pypi using twine.

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+
+
+Week 11: August 5 - August 11
+
+
+	- Install and test the newly uploaded package into virtualenv with different version of Python
```
jaylett, line 379: Which versions of Python do you propose supporting?

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+	- Test and fix bugs (if any)
+	- Discuss the credentials and register at pypi.org for twine upload as the uploaded generation archive remains temporary
```
jaylett, line 380: Again, I'm not sure why this is a significant amount of work. Going from code to uploaded pypi package is generally under a day's work following a tutorial. (The testing is of course going to take longer.)

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+
+
+Week 12: August 12 - August 18
+
+
+	- Finally, execution of "pip install xapian" into the virtualenv with different versions
```
jaylett: What is different between this and week 11?

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+	- Documentation of PyPI package
```
jaylett: What is different between this and the README in week 10?

proposal.rst
```diff
@@ -239,7 +264,113 @@ Project Timeline
...
+
+
+Final Evaluations: August 19 - August 26
+
+	- Final reading of documentation and discussion with mentor for any further changes
+	- Final code submission for evaluations
+
+Stretch Goals:
+
+	- Include Xapian-core in the PyPI package
```
jaylett, line 400: How important do you think this is for people coming to Xapian from the python world?

jaylett (commented):
I think this needs some more work. For instance, I'm not sure you have a clear view on the work required for pypy support. (Note in particular that adding a new language to SWIG is often a full GSoC project or more all by itself.)

ojwb:

PR #11, opened 2019-04-08
GSOC Draft proposal

PrithiPal:
Hello, 

This is my first draft. So please do not mind any grammatical mistakes. I expect if you can give me review on the topics I selected , or more specificity needed for explanation.

Thanks.

### DIFF
--- a/proposal.rst
+++ b/proposal.rst
@@ -82,7 +82,7 @@ Hello to you people reading this. I am Prithi, 5th BSc. Computer science undergr
 
 In school, I also hold onto two different research assistant positions. One is Bioinformatics based where I primarily wrote code for finding sequences in Proteins, producing statistics, and rough-pipeline for experimental/theoretical data comparisons. My second RA was based on Computational Linguistics where I focused on researching and implementing Evaluation metrics for LDA Topic-models based on Reddit data and wrote metric to quantify some real-time variables such as knowledge of user etc.  
 
-Now Apart from school and work, I participated in competition to create a drawing robot where I designed and implemented an interactive Image-processing/segmentation application on MATLAB. To be honest, that was one of the biggest project which took months to develop and our team placed first for robot which drew a photograph given on spot.
+Now Apart from school and work, I participated in competition to create a drawing robot where I designed and implemented an interactive Image-processing/segmentation application on MATLAB. To be honest, that was one of the biggest project which took months to develop and our team placed first for robot which drew a photograph given on spot. You can find our mention here : https://www.sfu.ca/computing/newsandevents/2017/inaugural-robotics-competition.html
 
 On the non-technical note, I started volunteering starting the second year and carried on front-end roles such as advisor, peer-mentor and workshop facilitation. 
 
--- a/proposal.rst
+++ b/proposal.rst
@@ -73,7 +73,7 @@ About You
 
  * IRC nickname(s): jugnu
 
- * Any personal websites, blogs, social media, etc: LinkedIn : www.linkedin.com/in/prithipal-285
+ * Any personal websites, blogs, social media, etc: LinkedIn : www.linkedin.com/in/prithipal-285(https://drive.google.com/open?id=1r3EilRwyBrFqlcEVYvSUWlI4jz8NN2Vb for partial pdf version of linkedin profile)
  * github URL: github.com/PrithiPal
 
  * Biography:
--- a/proposal.rst
+++ b/proposal.rst
@@ -119,6 +119,7 @@ COURSES :
 
 - NLP (project on context sensitive lexicons for sentiment analysis, assignments such as weighted-context-free-grammer, beam-search for scoring, HMM or vitebri, text-chunking model)
 - Research Assistant project on Evaluation methods for Topic Models (LDA-type). Read some papers on evaluation metrics, implemented them and also designed some metrics to quantify real-world variables from collected Reddit sub-thread data such as level of expertise of user towards the topic-being discussed.
+- Research Assistant project on sequence detection, profiling, producing statistics(graphs etc..) using Python/C++/Awk. 
 
 **What development platforms, tools and methods do you prefer to use?**
 
@@ -133,7 +134,7 @@ Some Useful Backends I know : Flask, Django, RabbitMQ, Celery, Scrapy, networkx.
 **Have you previously worked on a project of a similar scope?  If so, tell us
 about it.**
 
-At one of my internships, I worked on a data-pipeline for versioning of data-involved. The code-base involved lots of moving parts so the testing was dominated by Docker with multiple containers interacting. My strategy was to identify places with code change, change/test one file(if independent), test with few data examples and then scale for full testing on Docker (replicate the actual workflow including data-entering, Going through stages for processing, Storage, Sync with Kubernetes). 
+At one of my internships, I worked on a data-pipeline for versioning of data-involved. The code-base involved lots of moving parts so the testing was dominated by Docker with multiple containers interacting. My strategy was to identify places with code change, change/test one file(if independent), test with few data examples and then scale for full testing on Docker (replicate the actual workflow including data-entering, Going through stages for processing, BSON Storage, Sync with Kubernetes). Used Flask,Celery,RabbitMQ,Docker,OpenStack,Kubernetes. 
 
 
 **What timezone will you be in during the coding period?**
@@ -220,38 +221,36 @@ This component can be thought of as solution to the first three points in Projec
 		- Small Front end to illustrate numbers using charts and graphs. Store the results into files and import those files to make graph on them(jupyter notebook is a good starting point).
 		- If the previous is done, incorporate a small web-server(like a docker) which shows the interactive illustrations. For instance something like this one : https://pyldavis.readthedocs.io/en/latest/readme.html
 		- Cover all the code with the unit tests and component-testing. 
-
-At this point of time, the first option is more time consuming w.r.t just setting Omega up with templates. In the
 	
 
 **Do you have any preliminary findings or results which suggest that your
 approach is possible and likely to succeed?**
 
+- Consult with mentors before approach anything novel 
+- Document all my code including workflow to obtain same results.
+- In case of API, make unitests and then integrative tests if API interacts with multiple moving code-chunks.
+- For reporting system, I will implement algorithm through Cli, develop basic front-end, and then add more charts as required. 
 
-[DOCUMENTATION STEP] : 
-	- there are few chances that something will fail on massive scale. Actually the work can be divided into iteration such as "Practical Examples", "How to", "Core concepts". So for instance, I can develop and post user-friendly code for Practical examples, then how tos, then core-concepts. For the new user-friend API, 
-	 
-	- As of now, the project to support the front-end for Xapian is a big chunk of task. If however, something happens along the way, the "Rework Omega templates to use more modern web techniques" will ensure that we have some good front-end working. 
-
-	- Also for the benchmarking system, I can begin with a small prototype of front-end in python(such as )
+Overall in worst-case scenario, I may under-deliver but whatever deliver will have everything well-tested and documented.
 
 
 **What other approaches to have your considered, and why did you reject those in
 favour of your chosen approach?**
 
+- Earlier I thought of pursuing more analytical project (such as  Click-stream model, new weighting-schemes), however I decided in favour of documentation because I felt it took a considerable amount of time setting things up. I want to make other people be able to build, configure things in less time so that they can focus on functional stuff. So I put Documentation as first project component.
 
-
-FILLME
+- With the reporting system, because there are lot of breakable steps (obtain results (relevance judgment), make visualisations, scale to other ranking/scoring functions, data-set), I am sure this is more agile and not ending up in situation with "not know what to do". 
 
 **Please note any uncertainties or aspects which depend on further research or
 investigation.**
 
-FILLME
+- I would like to do more research on TracAdvanceSearchPlugin as I gave some buffer days in my Project timeline below.
+- Little more advanced data-profiling tools(for benchmarking visualization) even though I know how to make a basic one with a web-server.
 
 **How useful will your results be when not everything works out exactly as
 planned?**
 
-FILLME
+New developers starting with Xapian and people working on end-to-end LETOR api for easier model-testing and evaluation(to focus on actual algorithms instead of logistical issues).
 
 Project Timeline
 ----------------
@@ -373,7 +372,11 @@ Week 3
 
 Week 4
 	-  Buffer week if things are not completed from previous month otherwise continue.
-	
+
+MONTH 2 DELEVERABLE : 
+
+	[MINIMUM] : Started with Xapian backend support with TracAdvancedSearchPlugin and documented the procedure for future follow-ups.
+	[SATISFIABLE] : Fully integreate Xapian backend with TracAdvancedSearchPlugin and complete all the preceding projects.	
 
 Previous Discussion of your Project
 -----------------------------------
--- a/proposal.rst
+++ b/proposal.rst
@@ -60,9 +60,9 @@
 .. "How to write a kick-ass proposal for Google Summer of Code":
 .. https://teom.wordpress.com/2012/03/01/how-to-write-a-kick-ass-proposal-for-google-summer-of-code/
 
-======================================
-FILLME WITH THE TITLE OF YOUR PROPOSAL
-======================================
+====================================================
+Reporting system for easier algorithmic development
+====================================================
 
 About You
 =========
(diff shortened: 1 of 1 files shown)


proposal.rst
```diff
@@ -67,19 +67,22 @@ FILLME WITH THE TITLE OF YOUR PROPOSAL
...
- * Any personal websites, blogs, social media, etc: FILLME
-
- * github URL: FILLME
+ * Any personal websites, blogs, social media, etc: LinkedIn : www.linkedin.com/in/prithipal-285
+ * github URL: github.com/PrithiPal
 
  * Biography:
 
-.. Tell us a bit about yourself.
+Hello to you people reading this. I am Prithi, 5th BSc. Computer science undergrad @ Simon Fraser University, Burnaby, BC Canada. Talking about my academic experience, there are couple of courses that I enjoyed includes ones on Comparative Programming Languages (Go,Haskell,Clojure), NLP, Computer Vision, Compiler Design and why not Machine learning. On top of the school, I participated in two Co-ops with data-centric roles encompassing analytics and back-end building. In school, I also hold onto two different research assistant positions in Bioinformatics and Computational Linguistics for Reddit Sub-thread where I proposed and implemented solutions as per the requirements. I guess My linked will summarize stuff better.
```
ojwb: I'm really struggling to work out quite what this might mean:

> I also hold onto two different research assistant positions in Bioinformatics and Computational Linguistics for Reddit Sub-thread where I proposed and implemented solutions as per the requirements

Did two different sentences accidentally get spliced together?

Sadly linkedin seem to insist you create an account before you can view anything, so even if it does summarise better that doesn't really help me.

proposal.rst
```diff
@@ -96,53 +99,67 @@ Background Information
...
 systems for indexed text search.**
 
-FILLME
+I do not have prior experience with Xapian specifically. However I have rough understanding about indexing data-set/websites and how they serve behind search-engines. For instance, I created web-scraping programs while back using Python open source Scrapy framework and selenium to automate download-data from specific websites and store locally. Scrapy in particular have robust data-pipeline where user can fix things and prevent from re-inventing stuff. 
 
 **Tell us about any previous experience with Free Software and Open Source
 other than Xapian.**
 
-FILLME
+I have not contributed directly into a open-source codebase but I have utilised the open-source libraries/programs(python,matlab etc) to achieve tasks at hand.
```
ojwb: MATLAB is proprietary.

proposal.rst
```diff
@@ -152,43 +169,88 @@ Motivations
...
 **Who will benefit from your project and in what ways?**
 
 .. For example, think about the likely user-base, what they currently have to
 .. do and how your project will improve things for them.
 
-FILLME
+
+
+
+
```
ojwb, line 185: You haven't answered this question.

proposal.rst
```diff
@@ -152,43 +169,88 @@ Motivations
...
 
 Project Details
 ---------------
 
 .. Please go into plenty of detail in this section.
 
-**Describe any existing work and concepts on which your project is based.**
```
ojwb, line 169: You seem to have accidentally deleted this question.

proposal.rst
```diff
@@ -67,19 +67,22 @@ FILLME WITH THE TITLE OF YOUR PROPOSAL
 About You
```
ojwb, line 67: You haven't filled in the title a few lines above here.

proposal.rst
```diff
@@ -152,43 +169,88 @@ Motivations
...
 **Do you have any preliminary findings or results which suggest that your
 approach is possible and likely to succeed?**
 
-FILLME
+- Consult with mentors before approach anything novel 
+- Document all my code including workflow to obtain same results.
+- In case of API, make unitests and then integrative tests if API interacts with multiple moving code-chunks.
+- For reporting system, I will implement algorithm through Cli, develop basic front-end, and then add more charts as required. 
+
+Overall in worst-case scenario, I may under-deliver but whatever deliver will have everything well-tested and documented.
```
ojwb: None of this seems to actually answer the question being asked here though...

proposal.rst
```diff
@@ -239,7 +301,82 @@ Project Timeline
...
 .. in your project timeline.
 
-FILLME
+MONTH 1  : 
+
+BIG TASK NAME : DOCUMENTATION STEP
+
+Week 1 
+	- Read and understand code-base further. 
+	- Create a new VagrantFile and make sure that it installs all the xapian dependencies. I have a doc where I have all commands which helped me with my Xapian setup.
```
ojwb: Why do we need to create a new `VagrantFile`?

If [ceb80bbe06af6156f30e2553739b6f94a892f712](https://git.xapian.org/?p=xapian;a=commitdiff;h=ceb80bbe06af6156f30e2553739b6f94a892f712) didn't fully fix it, it's likely to be less work to iteratively fix what we have rather than throw everything out and start again.

proposal.rst
```diff
@@ -239,7 +301,82 @@ Project Timeline
...
+Week 3 
+	-  Buffer week if things are not completed from previous month otherwise continue.
+
+Week 4
+	-  Buffer week if things are not completed from previous month otherwise continue.
+
+MONTH 2 DELEVERABLE : 
+
+	[MINIMUM] : Started with Xapian backend support with TracAdvancedSearchPlugin and documented the procedure for future follow-ups.
+	[SATISFIABLE] : Fully integreate Xapian backend with TracAdvancedSearchPlugin and complete all the preceding projects.
```
ojwb: I assume this is actually meant to be "MONTH 3 DELIVERABLE"?

I'm concerned that you're trying to cram far too many different things into a 3 month project, without having actually yet dug into many of them to any great extent to get an idea for how much work they might actually involved - for example, I don't think we've seen any patches from you yet.

There's a common tendency for GSoC students to try to over-promise (perhaps in part thinking that might make their proposal look more appealing to the org, but I suspect mostly due to youthful optimism) but it's much better to start with a realistic project scope rather than have to start hacking off large chunks of the timeline when it becomes clear that it's not all going to happen.  This is why we tell you that you need to start to dig into the code that you'd working on, and to break down the tasks into subtasks until you get to tasks that you can actually usefully estimate the time for.  If one just thinks at a high level, one inevitably overlooks complicated details that will take a lot of the time.

I think you need to decide which part you want to work on most and focus on that.

proposal.rst
```diff
@@ -239,7 +301,82 @@ Project Timeline
...
+	- Update the "Installation" page with vagrant walk through tutorial.
+
+Week 2 
+	- Re-work on "Practical Example" tutorial to make it more user-friendly. 
+	- Add Omega documentation (in Xapian docs) for indexing, and how to obtain a front-end html(omindex, omega "query" [arguments])
+	- Start working on GSOC task:  "Rework Omega templates to use more modern web techniques"
+
+Week 3 
+	- Continue working on Omega templates.
+	- Identify areas that requires change user-end APIs, consult with mentors about this.
```
jaylett: Have you found anything where you'd recommend changing an existing API?

proposal.rst
```diff
@@ -239,7 +301,82 @@ Project Timeline
...
+
+BIG TASK NAME : DOCUMENTATION STEP
+
+Week 1 
+	- Read and understand code-base further. 
+	- Create a new VagrantFile and make sure that it installs all the xapian dependencies. I have a doc where I have all commands which helped me with my Xapian setup.
+	- Update the "Installation" page with vagrant walk through tutorial.
+
+Week 2 
+	- Re-work on "Practical Example" tutorial to make it more user-friendly.
```
jaylett: What sort of changes do you feel would make it better? (There is always room for improvement — what I'm asking is what aspects you think are unclear, and how you'd tackle them.)

proposal.rst
```diff
@@ -239,7 +301,82 @@ Project Timeline
...
+BIG TASK NAME : DOCUMENTATION STEP
+
+Week 1 
+	- Read and understand code-base further. 
+	- Create a new VagrantFile and make sure that it installs all the xapian dependencies. I have a doc where I have all commands which helped me with my Xapian setup.
+	- Update the "Installation" page with vagrant walk through tutorial.
+
+Week 2 
+	- Re-work on "Practical Example" tutorial to make it more user-friendly. 
+	- Add Omega documentation (in Xapian docs) for indexing, and how to obtain a front-end html(omindex, omega "query" [arguments])
```
jaylett: While I agree that Omega's documentation is very limited, GSoC is (sadly) expressly not about writing documentation. You can write documentation as part of delivering something else; writing example code that exercises Xapian or Omega would I think just about count. Given it's possible to get Omega running without writing new code, I'm wondering what the HTML frontend you'd implement and then document here would be, and how it'd be different to what's existing. (Again, the existing templates aren't great — as you've spotted, they could do with being moved to html5 at the least.)

proposal.rst
```diff
@@ -239,7 +301,82 @@ Project Timeline
...
+	- Start working on GSOC task:  "Rework Omega templates to use more modern web techniques"
+
+Week 3 
+	- Continue working on Omega templates.
+	- Identify areas that requires change user-end APIs, consult with mentors about this.
+	- Continue working on the user-end APIs.
+
+Week 4
+	- Continue working on the user-end APIs.
+	- Add tests to see consistency of new API().
```
jaylett: What sort of tests will highlight consistency problems with the API? Do you have examples (from Xapian or elsewhere) of the sorts of things you're looking for?

proposal.rst
```diff
@@ -239,7 +301,82 @@ Project Timeline
...
+MONTH 2  : 
+
+BIG TASK NAME : DOCUMENTATION STEP + REPORTING-SYSTEM
+DELEVARABLE OR SUBMITTABLE : 
+
+Week 1 
+	- Buffer week if things are not completed from previous month otherwise continue. 
+
+Week 2 
+	- Identify the requirements of a good reporting system(metrics, benchmarks, which algorithms etc..) as consulted by the developers on IRC and mentors
```
jaylett: Do you have an idea of which metrics and algorithms are relevant here? I'm not entirely clear from your proposal what this would look like. (Also, what are your measuring? Result relevance? Clustering? Impact of letor? All of the above?)

jaylett (commented):
I have the same concerns as Olly that you have a collection of things you want to work on — rather than a single coherent project. Almost any project can be extended, and a good one would enable you to discover future ideas as you implement earlier work.

proposal.rst
PrithiPal: Hey Jaylett, So on the "Introductions", I would recommend to merge the some parts from xapian-core/HACKING such as instructions on MacOSX and linux commands . Currently on Introductions page, there is not much code for demonstrating installation. Also for language specific bindings, I can borrow and take parts from xapian-bindings/INSTALL into main xapian docs installation page.

I found "Practical example" very good to follow, however I was more concerned about the "Core Concepts". For instance, here https://getting-started-with-xapian.readthedocs.io/en/latest/concepts/indexing/databases.html OR terms.html, there is lack of cli code for demonstration. I would definitely include some python examples for this. I understood the value concept but no know how it works in practice : https://getting-started-with-xapian.readthedocs.io/en/latest/concepts/indexing/values.html. So these are mere examples.

proposal.rst
PrithiPal: So first part will be little code examples of omindex and /bin/omega from https://xapian.org/docs/omega/quickstart.html. Also I followed this and received a html file(output of /usr/lib/omega/bin/omega 'P=my search terms' HITSPERPAGE=10 ), but the clicking search goes to page where it says "file not found". I do not have perfect understanding of cgi templates, yet however I would like to make that work and include in Xapian docs little intro to Omega. Also some coding examples which connects omega and xapian-delve(if I create db through xaipan-delve and omega reads from that database.)

So my steps would be : Get templates workable with cgi as suggested here : https://xapian.org/docs/omega/cgiparams.html, and then think about upgrading to more modern HTML5.

proposal.rst
PrithiPal: For instance in doc here : https://getting-started-with-xapian.readthedocs.io/en/latest/practical_example/indexing/writing_the_code.html, there's a command  : python2 code/python/index1.py data/100-objects-v1.csv db.

This is essentially calling a file and passing arguments to it. A user-end API would mean that I call this functionality something like xapian-docsprint --index [DATAFILE], or other flags such as --search, --delete,--search etc. I guess more can be found as I will start digging more inside.

proposal.rst
jaylett: We'd definitely expect you to get on top of CGI quite quickly — it's not complex and there are plenty of documents discussing it. (For instance, for Apache: https://httpd.apache.org/docs/2.4/howto/cgi.html — and indeed we have a worked example on our wiki, although it's for Xapian 1.0 so perhaps should be updated. I'd recommend working through that example and updating it if needed; it will give you a much better idea of what you might need to do for Omega documentation.)

The problem you're encountering, as I believe was pointed out on IRC (but you may have missed) is that you're trying to treat the HTML generated by a test run of omega as if it's a live site. If you set up omega to run as  CGI with a web server, you shouldn't have this problem.

proposal.rst
PrithiPal: Okay, according to my understanding, there are two sets of algorithms : xapian-core/weights (which are based on weight-schemes) and others in xapian-letor/ranker | scorer. To obtain the results, there are nice tests which can be good starting points : xapian-core/tests/api_backend.cc and xapian-letor/tests/api_letor.cc. As a first step, I will go through and obtain results (relevant-judgments) for instance DEFINE_TESTCASE(svm_ranker_one_file, generated) which does for SVM ranker algorithm. Then I plan to create a python-api like this : xapian-weight [algorithm] [dataset] or xapian-ranker [ranking-algorithm] [dataset] [evaluation], and depending upon the complexities add flags for further options. 

As breakable steps, good one will be make python-Api for weight-scheme, test them (like unit-tests), and maybe add to Xapian docs for examples like here : https://getting-started-with-xapian.readthedocs.io/en/latest/howtos/weighting_scheme.html or incorporate this: https://github.com/xapian/xapian/blob/master/xapian-letor/docs/letor.rst (with python-api) into Xapian docs.

For actual measurments, good place to start is to have score benchmarks like here : https://trac.xapian.org/wiki/GSoC2011/LTR/Notes#IREvaluationofLetorrankingscheme. Generally, the comparison purely depends upon algorithms. This example is for LETOR algorithms and may/maynot apply for weighting-schemes(because they are unsupervised algorithms). 

Cluster algorithms in xapian-core/cluster/* are another set of algorithms. However I have to read more about the input-dataset, the output(metric) and see the numbers which can be compared. 

So types of algorithms = LETOR-BASED, WEIGHT-SCHEMES, CLUSTERS and probably other If I missed.

proposal.rst
PrithiPal: I will fix the writing in my next commit. Thanks.

proposal.rst
PrithiPal: I will fix this in my next commit. Thanks

proposal.rst
PrithiPal: I will fix this in my next commit. Thanks

proposal.rst
PrithiPal: I will fix this in my next commit. Thanks.

proposal.rst
PrithiPal: I didn't noticed the new ticket based on Vagrant. I will look into this. Thanks and try to re-build xapian on my system.

proposal.rst
PrithiPal: I will fix the writing in my next commit. Thanks.

proposal.rst
PrithiPal: So these unit-tests may include the python-api that I would have created for wrapping around algorithms (discussed in below conversations) as well as the that wraps around xapian-docsprint/code/python. As of now, I do not think if there is any involvement of build system or any other systems, so these tests should be fairly simple like testing individual functions.

proposal.rst
PrithiPal: Thanks Olly for pointing this out. I am kind of feeling the same, that Trac is another pile of problem. I am planning to spend my day today to go through to setup on Trac's page : https://trac.edgewall.org/wiki/TracDev. If I product that on my system, I will surely add more details about the specifics of starting workable steps to Add Xapian backend on this front-end plugin. 

Otherwise if things not workout till tomorrow, I will spread out my objectives on only focus on DOCUMENTATION and REPORTING-SYSTEM. 

In my understanding there are TWO front-ends I have in my mind. One is the basic for REPORTING-SYSTEM for algorithm benchmarking(speed,relative evaluation) metrics, and then second is front-end for indexer (Omega currently). If TracApplicationSearchPlugin is implemented, then that would be a layer on Omega CGI(present basic front-end).

proposal.rst
PrithiPal: Thanks for the note. I missed from the IRC coversation. I will start working on this.

proposal.rst
ojwb: The instructions in `xapian-core/HACKING` are for people fetching the code from git with the intention of doing development work on Xapian itself, not for people installing releases of Xapian and just wanting to use it.  The instructions for installing releases are in `xapian-core/INSTALL`.  These are aimed at different audiences, and the former instructions are significantly more complicated so it's not helpful to mix them.

But most people just wanting to use Xapian would not build from source but instead install a package of it through their distribution's package manager.

ojwb:

PrithiPal (commented):
This is the latest proposal.

proposal.rst
ojwb: I'm not sure I understand what wrapping `xapian-docsprint/code/python` achieves.  That seems like it would just obscure the code in there which actually does the work, and that code exists solely as examples to demonstrate how to use particular API features, so clarity is a high priority.

Elsewhere you say about this:
> A user-end API would mean that I call this functionality something like xapian-docsprint --index [DATAFILE], or other flags such as --search, --delete,--search etc.

And how is that useful?

PrithiPal: This is my old proposal. I have made significant changes to my latest proposal where I concentrated more on reported Systems rather than these documentations.

ojwb: OK - it showed as a new comment from you after you asked me to look at your proposal again.

proposal.rst
```diff
@@ -273,11 +399,31 @@ Use of Existing Code
...
+- letor algorithm calling API : xapian-letor/bin/xapian-train and xapian-letor/bin/xapian-rank.cc
+- letor algorithm workflow example : xapian-letor/tests/api_letor.cc 
+- letor algorithm evaluation : xapian-letor/ranker. 
+- letor docs : xapian-letor/docs
+
+- Omega templates : xapian-applications/omega/templates..
+- Stemming data : xapian-data/stemming
+
+
+
```
ojwb, line 434: This isn't asking about code that's already in Xapian - see comment in the template below for the motivation for this question.

proposal.rst
```diff
@@ -239,7 +308,64 @@ Project Timeline
...
 
-FILLME
+**Month 1 Theme** : *DOCUMENTATION AND OMEGA *
+
+**Week 1** 
+	- Read and understand code-base further. 
+	- Identify the families of algorithms inside the code. Currently known to me are weighting-scheme, letor, spelling-checking, cluster, stemming
+
+**Week 2** 
+	- solve "Rework Omega templates to use more modern web techniques"
```
ojwb, line 318: Above you said you plan to do this before the GSoC results are out.

proposal.rst
```diff
@@ -96,99 +101,163 @@ Background Information
...
 
 Project Details
 ---------------
 
+**Describe any existing work and concepts on which your project is based.**
+
 .. Please go into plenty of detail in this section.
 
-**Describe any existing work and concepts on which your project is based.**
+I can break down my project into two big components : Reporting system for algorithm development and testing and support for TracApplicationSoftwarePlugin.
```
ojwb, line 194: Are you still actually planning to work on TracApplicationSofwarePlugin?  You seem to have removed all the other references to it but left this one.

proposal.rst
PrithiPal: Yes. I will start tackling, probably starting this weekend. However, still I wrote that in Initial week schedule in case some delay happens.

proposal.rst
PrithiPal: Nopes. I have made few more revisions and uploaded to my fork as well as Google submission page. So to make things clear , I am concentrate on little documentation parts + reporting system. Reporting system is more about getting different algoirhtms together on one platform like a CLI/system to visualize evaluative metrics. However studying specifics will take time and I try best to update my proposal even after Google submission.

proposal.rst
PrithiPal: I see that. I have updated my proposal with the external libraries tool I am looking to incoporate. Thanks for the mention.

proposal.rst
ojwb: > I try best to update my proposal even after Google submission.

Sorry, but we can't take updates you make after the deadline into account - Google explicitly tell us we can't:

_Your decision about whether or not to select a student should be based on the Final PDF proposal. You will also be able to reference the draft proposal -- but should not make your decision based on it. A draft proposal is intended to be used to view the comment thread and maybe see how the student interacted with you and responded to comments, etc. Do not use it to let the student make more changes after the deadline. That is not fair to all the other applicants._

PrithiPal: I understand, I meant to say that I will update that for my own convinience and documentation and not as a submittable document after the deadline.

ojwb:

PR #12, opened 2019-04-08
[GSoC 2019] Improve Estimated Total Number of Results Proposal (@Teng Ma)

stmatengss:
Appreciate for your reviews.
Many Thanks.


proposal.rst
```diff
@@ -96,53 +95,61 @@ Background Information
...
 **What timezone will you be in during the coding period?**
 
-FILLME
+UTC+8
 
 **Will your Summer of Code project be the main focus of your time during the
 program?**
 
-FILLME
+Yes, this project will be treated as the primary one except some unresistible factors.
```
ojwb, line 138: What are these "unresistible factors"?

It's not necessarily a problem if there are some small existing commitments, but we really need to be clear on what they are.

proposal.rst
```diff
@@ -168,82 +173,124 @@ Project Details
 
 **Describe any existing work and concepts on which your project is based.**
 
-FILLME
+First, I have collected several resources about how to implement and improve the estimation.
+
+*** Academic
+1. (Estimating Corpus Size via Queries [CIKM'06])[http://fontoura.org/papers/cikm2006.pdf] This paper uses the basic estimator estimate the search result size. The first approach requires a uniform random sample of documents from the corpus. It avoids this notoriously difficult sample generation problem, and instead uses two fairly uncorrelated sets of terms as query pools; the accuracy of the second approach depends on the degree of correlation among the two sets of terms.
```
ojwb: This paper is describing how to estimate the total number of documents in the collection when you can only run queries and see how many documents match (for example, you could use this to estimate how many pages Google index).

That's a related problem to the one we want to solve, but it seems to me to be essentially a different problem.  We already exactly know the collection size, but when running a query we want to try to return a better estimate of the number of matches.  If we would need to run extra queries to improve the estimate we might as well just consider additional documents for the query we're already running.

proposal.rst
```diff
@@ -168,82 +173,124 @@ Project Details
...
+2. (Random Sampling from a Search Engine’s Index [WWW'06]) [http://www2006.org/programme/files/pdf/3047.pdf] This paper proposes a lexicon-based technique and a random
+walk technique to avoid long documents and highly ranked documents.
```
ojwb: This paper is about a technique to sample documents randomly from a search engines index when you can only run queries.

If we wanted to sample documents randomly in Xapian we don't need to go to such lengths.  In the case where the docids are contiguous it's very easy (and we can tell if this is the case).  When there are gaps we could just pick a random docid in the range that's used and resample if we get an unused one, though maybe there's a more sophisticated approach.  But even this is going to be much more efficient than running queries to sample documents.

proposal.rst
```diff
@@ -168,82 +173,124 @@ Project Details
...
+3. (Efficient Search Engine Measurements [WWW'07])[http://wwwconference.org/www2007/papers/paper753.pdf] This paper present two estimators that are able to remedy the skew introduced by approximate degrees. This estimators are based on a careful implementation of an approximate importance sampling procedure.
```
ojwb: And this paper is similarly about "external evaluation of search engines".

ojwb:
The big thing I'm lacking is a clear idea of what you actually proposing to implement here.  You've referenced three papers which are somewhat related, but none of them actually seem to provide techniques which can be directly applied here.

Perhaps some sort of sampling or random walk of matching documents can be done, but I'd really like to have a clearer idea of what that might look like.  There seems to be a significant risk that such an approach could incur more overhead than the existing crude solution we have to improving the estimate.  If I actually knew what's being proposed I could think about what that would entail.

proposal.rst
stmatengss: I have the lumbar disc herniation (LDH). Sometimes, I have to go to my doctor. This one is the main unresistible factors.

proposal.rst
stmatengss: Indeed, this paper is an evaluate paper. However, it gives some hints about how to remedy the bias phenomenon while estimating. In this project, I guess the core contribution is to eliminate the skewed results since the words from a query are usually correlated.

stmatengss:
> The big thing I'm lacking is a clear idea of what you actually proposing to implement here. You've referenced three papers which are somewhat related, but none of them actually seem to provide techniques which can be directly applied here.
> 
> Perhaps some sort of sampling or random walk of matching documents can be done, but I'd really like to have a clearer idea of what that might look like. There seems to be a significant risk that such an approach could incur more overhead than the existing crude solution we have to improving the estimate. If I actually knew what's being proposed I could think about what that would entail.

Sorry for my misleading. On my mind, the result for every single term is certain, so I think a rough estimation of results is easy to implement by using probability theory (just like sum(set 1...n)). However, what I want to do better is to eliminate the bias between the lower-bound and upper-bound estimation and finally produce an approximate result to reach a correct result as far as possible. So I guess I should spend another one or two weeks to know this area more deeply. For me, a first step is to try random walk or sampling, but until now, I didn't have any evidence to show it is effective. Can you give more some detailed suggestions? Thank You~

proposal.rst
stmatengss: Got it.

stmatengss:
Another question: I worry about some of the effects can only be validated after testing real-world dataset. Such dataset will be easy to acquire from the internet? Or do we have related private dataset?

ojwb:
> On my mind, the result for every single term is certain, so I think a rough estimation of results is easy to implement by using probability theory

Yes, and we already do that.

> For me, a first step is to try random walk or sampling, but until now, I didn't have any evidence to show it is effective.

It would be good to demonstrate how effective it can be, but it is a fairly small leap of faith to assume it would be in many cases.  My bigger concern is how much extra work it would need (especially I/O when we don't have everything we need cached).

> However, what I want to do better is to eliminate the bias between the lower-bound and upper-bound estimation and finally produce an approximate result to reach a correct result as far as possible. So I guess I should spend another one or two weeks to know this area more deeply.

Your problem here is that you only have about one day to the submission deadline, and Google explicitly tell us we should judge the final submitted proposal and not allow additional updates outside their system.  You can work on patches, etc after the submission deadline, but it sounds like you're talking about a week or two of research before you finalise your proposal rather than a week or two to come up with some sort of prototype.

> Can you give more some detailed suggestions?

I think the suggestion in the project idea to make use of the document id range information for each term would help significantly for very little overhead (and probably no extra I/O).  For example, if `foo` only occurs in the docid range [10,20] and `bar` in [18, 30] then we know that `foo AND bar` can only match in the range [18, 20].

We might also be able to apply a sampling-type approach to the documents we are already considering.  It isn't a uniformly random sample (it will tend to be skewed towards lower document ids because the bar to picking a document tends to rise as we work through the matching process in ascending docid order) but it wouldn't require extra I/O.  And perhaps it's even possible to compensate for this skew to some extent.

> Such dataset will be easy to acquire from the internet? Or do we have related private dataset?

It's hard to find public datasets with corresponding queries - attempts to release query logs have proved problematic for privacy reasons (e.g. https://en.wikipedia.org/wiki/AOL_search_data_leak was a intentional release of query logs, but lead to the AOL CTO and two researchers losing their jobs).

There are some non-public research datasets which it's possible to arrange access to.  Currently we have access to FIRE, and we're working on access to another one.

Another option might be to download a wikipedia dump for the documents, and make up a list of queries (you could probably have a look at your own search history and pick out non-sensitive queries that make sense as a search in wikipedia).  That's a bit more artificial, but for this project that's probably OK.

ojwb:
BTW, I wouldn't delay filing a "final" proposal until the last minute.  You can upload a revised "final" proposal as many times as you like up to the deadline, and you don't want to miss the deadline due to an internet problem or power outage.

stmatengss:
> BTW, I wouldn't delay filing a "final" proposal until the last minute. You can upload a revised "final" proposal as many times as you like up to the deadline, and you don't want to miss the deadline due to an internet problem or power outage.

Thank you for your reminder. All these comments are extremely useful to me. I will improve my proposal as soon as possible.

stmatengss:
Another question: Do I need to improve my time schedule?  Or my planned time is enough for this project, right?

stmatengss:
A Minimal Variance Estimator for the Cardinality of Big Data Set Intersection [SigKDD 17]
This paper gives a maximum likelihood method to estimate the Intersection between two sets with a low IO overhead (streaming). We can gain the result through `set1 + set2 - the Intersection of set1 and set2`
A guess this paper is closely related to this project.

stmatengss:
Hey all,
I have made some changes according to the comments as another commit.
Thank you all for careful review.

ojwb:

PR #13, opened 2019-04-08
Update proposal.rst

gayatripalkar:
https://github.com/xapian/xapian-gsoc-proposal/commits?author=gayatripalkar

### DIFF
--- a/proposal.rst
+++ b/proposal.rst
@@ -67,23 +67,22 @@ FILLME WITH THE TITLE OF YOUR PROPOSAL
 About You
 =========
 
- * Name: FILLME
+ * Name: Gayatri Palkar
 
- * E-mail address: FILLME
+ * E-mail address: gayatripalkarhere@gmail.com
 
- * IRC nickname(s): FILLME
+ * IRC nickname(s): gayatri_palkar
 
- * Any personal websites, blogs, social media, etc: FILLME
+ * Any personal websites, blogs, social media, etc: My Linkedin Profile: https://www.linkedin.com/in/gayatri-palkar-9a3509180 
 
- * github URL: FILLME
+ * github handle: gayatripalkar
 
  * Biography:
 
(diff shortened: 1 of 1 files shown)


proposal.rst
```diff
@@ -67,23 +67,22 @@ FILLME WITH THE TITLE OF YOUR PROPOSAL
 About You
```
jaylett, line 67: You've missed setting the title of your proposal, a little above here.

jaylett (commented):
You need to provide quite a bit more detail. For instance, we expect a timeline with an idea of what you'd do on a week by week basis. You've also linked to a _group_ of project ideas, but not indicated which one or ones you want to tackle — or how you'll tackle them. We have some notes on how to write a good timeline (and a good proposal) which may help here.

proposal.rst
```diff
@@ -152,43 +154,51 @@ Motivations
...
 .. For example, think about the likely user-base, what they currently have to
 .. do and how your project will improve things for them.
 
-FILLME
+My project would benefit the majority of people that spend a lot of time being online. It would also help students
+
+Project Details:
+
+Idea Title: Projects related to testing
+Idea Link: https://trac.xapian.org/wiki/GSoCProjectIdeas#ProjectsRelatedtoTesting
```
ojwb, line 173: A proposal needs to be much more than just some details about yourself and a link to a project idea (or worse a group of project ideas).

You need to decide which project idea from that group you want to do, then dig into what implementing the project would involve, start to work out how you'd implement it, and show us that you understand what you're proposing to do.

ojwb:
I notice you don't seem to have filed a "final" proposal yet - please don't leave this until the last minute. You can upload a revised "final" proposal as many times as you like up to the deadline, and you don't want to miss the deadline due to an internet problem or power outage.  Google are very strict about not giving extensions.

ojwb:

PR #14, opened 2019-04-09
GSoC Proposal

iamhks:
Please give me a feedback on my proposal.

### DIFF
--- a/proposal.rst
+++ b/proposal.rst
@@ -1,5 +1,5 @@
 .. This document is written in reStructuredText, a simple and unobstrusive
-.. markup language.  For an introduction to reStructuredText see:
+.. markup language.  For an introductiont to reStructuredText see:
 .. 
 .. https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html
 .. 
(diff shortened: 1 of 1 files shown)


proposal.rst
```diff
@@ -168,27 +171,27 @@ Project Details
...
 
-FILLME
+The approach was suggested in the ideas page and it seems perfectly fine. And the possibility of success is very high and the things that could possible go wrong are pretty low.
 
-**What other approaches to have your considered, and why did you reject those in
+**What other approaches have you considered, and why did you reject those in
 favour of your chosen approach?**
 
-FILLME
+I was totally fine with the approach suggested in the ideas page until this question was asked. So, what I thought was to declare and initialize a new variable and keep incrementing and displaying the results as per the user requirement (an instance of 10 in the first page) but this approach is not efficient since it will work only when the searched item is single word. But when we have and array of strings containing more than a word this approach will fail and hence I rejected it. So, I will be sticking with the approach suggested by the ideas page and as elaborated in the previous to previous question.
```
ojwb: When the search query is a single word the estimated number of matches is already exact (because it's just the number of different documents which that term occurs, which we have stored for each term).

proposal.rst
```diff
@@ -239,7 +242,37 @@ Project Timeline
...
+
+Week 9: Second Evalutaion Week- will review the code and documentation and after discussing with mentor will submit the evalutaion.
+
+Week 10: Code part 3 of the algorithm 
+
+Week 11: Error testing and debugging
+
+Week 12: Discuss with mentor and document the entire progress in detail.
+
+Final Evalutaion- After the completion of the project, I will look into it once more to find any bugs or errors, will also go through the documentation, and approve with my mentor before the submission of my project.
```
ojwb: This is all written in incredibly vague terms - almost nothing here is usefully measurable.

At each evaluation the mentor has to say if you're behind schedule, on schedule, or ahead of schedule, but since none of the tasks are clearly described it would be pretty much impossible to say.  For example, I can't tell if you've completed "Code part 1 of the algorithm" when you haven't defined anywhere what "part 1 of the algorithm" actually is.

You need to dig in to the project you're proposing and the areas of the code you'll be working on to get an understanding for what is actually involved in what you are proposing to do.  Without you having done that, it's hard for you to write a realistic proposal for the project.

You also appear to be only proposing to tackle the part of the project idea about using the docid ranges, and not the part about running some experiments to try to tune the estimates.  You're welcome to take a project idea and adjust it by adding parts and dropping others (though the project still needs to make sense as a whole so it's a good idea to discuss with us if you plan to do this).  By taking just half of a project I suspect this probably ends up being too little work to fill 3 months (it's hard to judge though, as you haven't submitted a patch, so I don't have any tangible evidence of your coding abilities).

ojwb:

PR #16, opened 2020-03-18
 Improve estimated total number of results

bodasadallah:


proposal.rst
```diff
@@ -38,56 +38,64 @@
...
  * Biography:
 
-.. Tell us a bit about yourself.
 
-FILLME
+
+	Hey, I am Boda, Computer Engineering student from Egypt.
+	I have a great passion for Competitve programming, I have praticipated in many
+	`ACM  <https://en.wikipedia.org/wiki/International_Collegiate_Programming_Contest/>`_ contests
+	Last year we came in third in local contest and qualified to next round. I praticipated in Google Hash Code this year, and came in #31 on Egypt.
```
jaylett: If you look at the rendered output, this block is over-indented. (Minor point!)

proposal.rst
```diff
@@ -38,56 +38,64 @@
...
+ * github URL: https://github.com/BodaSadalla98
 
  * Biography:
 
-.. Tell us a bit about yourself.
 
-FILLME
+
+	Hey, I am Boda, Computer Engineering student from Egypt.
+	I have a great passion for Competitve programming, I have praticipated in many
```
ojwb: You might want to spell-check this (e.g. "Competitve" and "praticipated" are both typos)

proposal.rst
```diff
@@ -197,45 +233,88 @@ Project Details
...
 
 **Do you have any preliminary findings or results which suggest that your
 approach is possible and likely to succeed?**
 
-FILLME
+There's approach addressed in a `patch <https://oligarchy.co.uk/xapian/patches/docid-ranges-in-matcher.patch/>`_ by Olly,
+that make use of the fact that we have the ids for the first and last documents now.
+Example: for the "MultiAndPostList", we would mutiply the probability of each term independantly, and we would do that by
+divide the term frequnecy(from term_freq_est() function) by the database size. But now we can get the range from the first docid to the last docid
+(last - first +1), and we use that range instead of the whole database size.
```
ojwb, line 245: So AND and OR are mostly done already, but what does the calculation look like for some of the other operators?  For example `AND_NOT`?

proposal.rst
```diff
@@ -197,45 +233,88 @@ Project Details
 
 **Describe any existing work and concepts on which your project is based.**
 
-FILLME
+
+I am still trying to see if we can utilize the N-gram algorithm to help on this task.I made a google `doc <https://docs.google.com/document/d/1yXdnAhtNKWcODLLBy2hB2pluQ2nr3EfZJP2_F1XIpDs/edit
+/>`_ that I keep updated with results I reached so far.
```
jaylett: This link doesn't go anywhere. In any case, when you submit your final proposal it should contain the details of what you're proposing to do; you shouldn't link to a separate document that you can change after the submission deadline.

proposal.rst
```diff
@@ -197,45 +233,88 @@ Project Details
...
+
+
+10 June : Final exams (this is an estimate,official schedule isn't announced yet.)
+
+
+* Two Weeks: Working on OP_FILTER, OP_NEAR, and OP_PHRASE (Code, testing, and doumentation.)
+
+* Two Weeks: Working on OP_VALUE_RANGE, OP_VALUE_LE, OP_VALUE_LT, OP_VALUE_GE, and OP_VAUE_GT (Code, testing, and doumentation).
+
+* Two Weeks: Working on OP_SCALE_WEIGHT,and OP_ELITE_LIST (Code, testing, and doumentation.)
```
jaylett: You haven't followed our guidance about breaking things down to a week _at most_. Additionally, I can't tell from your proposal at the moment anything concrete about what you'll be doing in any of these periods. I'd like to see a discussion of the background to the approach overall, so I understand what work is needed specifically for OP_FILTER, for instance.

In particular, we suggest in our guidance that your project should be made up of several smaller projects, each of which can be completed and merged and which will produce some improvement for Xapian. I don't understand at the moment where this work could be merged, or what it would do. For instance, if we merged it at this point, what are we going to get next out of the work on OP_SYNONYM and OP_MAX?

proposal.rst
```diff
@@ -197,45 +233,88 @@ Project Details
...
+
+* One Weeks: Working on OP_SYNONYM,and OP_MAX (Code, testing, and doumentation.)
+
+* One Weeks: Working on OP_WILDCARD,and OP_INVALID (Code, testing, and doumentation.)
+
+This could be done in a shorter amount of time. If so, i can start work on other projects.
```
jaylett: What other projects? It's a good idea to present some "stretch goals" for a project if you believe you'll complete it before the end of the programme. (These don't have to be an extension of the earlier work if there isn't anything that makes sense.)

proposal.rst
```diff
@@ -102,48 +110,73 @@ Background Information
...
+
 
 **Will your Summer of Code project be the main focus of your time during the
 program?**
 
+Yes.
+
+
+
+
```
jaylett: This PR isn't as important as others, but it's a good idea to get into good habits. These additional lines shouldn't really be there. If you create commits by staging with confirmation (git add -p) then you'll see the diff hunks before you make the commits, and this will be easier to spot and avoid.

jaylett (commented):
You need to spend more time breaking down the project, and also explaining how you're approaching the problem. Your final proposal needs to stand on its own as much as possible, so you should explain what your changes are attempting to do (ie why they'll improve the estimate) at each step.

proposal.rst
ojwb: Indeed - one of the reasons for us using restructured text in git and github PRs for proposal review is that these are tools that you'd need to use during your project (definitely for git and github PRs, and very likely for restructured text), so it helps you get a bit more familiar with these tools, and helps us see that you can use them appropriately.

proposal.rst
```diff
@@ -262,7 +276,8 @@ and take huge space.
 **Please note any uncertainties or aspects which depend on further research or
 investigation.**
 
-I am still not sure if we can come up with better approach. Or how better the result would be for approach using the first and last doc ids
+We still assume independance in calculating the estimates.
```
ojwb: Well, yes and no.  Currently we always just assume the terms occur independently.

Using the docid range information effectively feeds some information about dependencies between terms into the calculation.  In some cases that gives us perfect information (for example, in the degenerate case where the terms have disjoint ranges of docids we can tell that they don't occur together) whereas in others we get no useful information and have to assume independence (for example, if the terms both occur in the first and in the last used docids we'd just use the same estimate as currently).

So we do still assume independence in some cases, but that's an improvement on always assuming it.

proposal.rst
```diff
@@ -272,34 +287,138 @@ It would still produce better estimates as we reduced the number of documents we
 Project Timeline
 ----------------
 
-* Three Weeks of Bonding:
+The main goal of this project is to improve the estimation of the total number of results buy making use of the
+the new feature that gives the first and last docids for each term.
+
+The task is to improve the Xapian::MSet::get_matches_estimated() method for 17 operators, but OP_AND and OP_OR are almost done, so that leave us with 15 operators.
```
ojwb: As I said on chat, I think you want to consider the `PostList` subclasses rather than the operators.  In some cases they are pretty much equivalent, but not in all cases.

proposal.rst
```diff
@@ -272,34 +287,138 @@ It would still produce better estimates as we reduced the number of documents we
...
+Second Evaluation
 
-* One Weeks: Working on OP_SYNONYM,and OP_MAX (Code, testing, and doumentation.)
+week 9 : Work on OP_VALUE_RANGE
 
-* One Weeks: Working on OP_WILDCARD,and OP_INVALID (Code, testing, and doumentation.)
+week 10 : Work on OP_VALUE_GT, OP_VALUE_GE, OP_VALUE_LT, and OP_VALUE_LE
 
-This could be done in a shorter amount of time. If so, i can start work on other projects.
+week 11 , 12 : Review, see if anything need to change.
```
ojwb: That's rather a nebulous task for two weeks.

It's good to allow for things potentially overrunning, but a better way to handle this is to identify some "stretch goals" which you'd do if time allows, but would be skipped if things overrun.

The project idea even suggests something else that could be looked at:

> if we ran some experiments using real datasets and associated queries we could look at how the estimates compare to the true answers, and use the results to tune the calculations of the estimate to try to reduce the skewing effect we get from assuming independence.

But you're welcome to think up useful things for yourself - for example, your idea of storing some sort of co-occurrence data from a standard dataset (a challenge there seems to be storing it compactly enough for the cost not to be too high for the benefit, but perhaps we can just store data for cases which differ much from independence or something like that).

proposal.rst
```diff
@@ -327,16 +335,18 @@ FILTER :
 
 	then we just return probability of L times the probability of R times the total range, which would be the union of the two ranges.
 
-NEAR, and PHRASE :
+NEAR(nearpostlist), PHRASE(phrasepostlist), and exactphrasepostlist :
 
-	No work needed on this.
+	implementing get_used_docid_range().
+
+	I think we can treat it like MULTI_AND, and the divide that estimate by constant like what we doing now.
```
ojwb: The way these 3 work is the same.  Considering NearPostList, this just performs the positional check, taking a stream of postings to check from a sub-postlist object.  The subqueries of the OP_NEAR are combined with other adjacent "and-like" queries, and the positional check is hoisted as high as we can, so that it has to check fewer documents.

So e.g. `(a AND b) AND (c NEAR d)` -> `NearPostList(MultiAndPostList(a, b, c, d))` (and the NearPostList knows that it should check the positions of `c` and `d`.

I think this means that most (maybe all) of the work of using the ranges is done for you by the `MultiAndPostList` inside the `NearPostList`.

proposal.rst
```diff
@@ -337,9 +337,13 @@ FILTER :
...
 
-	implementing get_used_docid_range().
+	Most of the work done by MultiAndPostList.
+
+	I think we can do better than just divide the estimate number by constant.
+
+	We can get a list by words that most likely to come next to each other, like: "the" + "noun".
 
-	I think we can treat it like MULTI_AND, and the divide that estimate by constant like what we doing now.
+	This would help most in exactphrasepostlist, where we search for words that make a sentence (terms right next to each other)
```
ojwb: Where would you get this list of words that are likely to be adjacent?

How would you store it?

And how would you apply it to the non-tight phrase and near cases?

Dividing by an arbitrary constant certainly isn't a great approach, but I think you need a clearer plan for how to better solve this.

ojwb:

bodasadallah:
Thank you so much, for your feedback. I believe I did submit a final proposal there, my name is "Abdelrahman Atef Mohamed"

ojwb:
> I believe I did submit a final proposal there

Oh yes, I see that - I'd just failed to match that with your github username.

jaylett:
@BodaSadalla98 if you haven't uploaded your final proposal, you need to do that in the next 40 minutes.

bodasadallah:
@jaylett  Thank you so much for your concern, It should be uploaded  with last changes I made.

ojwb:
Thanks for you proposal.

PR #17, opened 2020-03-21
GSoC Proposal for project - Implement Bindings in Go

srinivasyadav18:
Please review it. Im Happy to receive any kind of feedback . Thank You 😊

### DIFF
--- a/proposal.rst
+++ b/proposal.rst
@@ -338,18 +338,28 @@ Plan :
         code :
  
             func main(){
+
             tm := xapian.NewTermGenerator()
+
             // tm.Set_stemming_strategy(xapian.TermGeneratorSTEM_NONE) --> fails
+
     
             // for Stem Startegy enum in TermGenerator class
+
             // swig defines a type XapianTermGeneratorStem_strategy int
+
             // and for each element in  enum a type is created as
+
             // type TermGeneratorSTEM_SOME int is created.
+
             // so before passing to the functions converion should happen
+
             // even both hold the same internal type.
           
             tm.Set_stemming_strategy(xapian.TermGeneratorSTEM_NONE(xapian.TermGeneratorSTEM_SOME))
+
             fmt.Println(tm)
+            
             }
  
         OUTPUT(WHEN FAILED) :
--- a/proposal.rst
+++ b/proposal.rst
@@ -67,27 +67,28 @@
 .. https://teom.wordpress.com/2012/03/01/how-to-write-a-kick-ass-proposal-for-google-summer-of-code/
 
 ======================================
-FILLME WITH THE TITLE OF YOUR PROPOSAL
+GSOC 2020 - PROPOSAL
 ======================================
 
 About You
 =========
 
- * Name: FILLME
+ * Name: S SRINIVAS YADAV
 
- * E-mail address: FILLME
+ * E-mail address: vasu.srinivasvasu.14@gmail.com
 
- * IRC nickname(s): FILLME
+ * IRC nickname(s): srinivasyadav227
 
- * Any personal websites, blogs, social media, etc: FILLME
+ * Any personal websites, blogs, social media, etc: NIL
 
- * github URL: FILLME
+ * github URL: https://github.com/srinivasyadav18
 
  * Biography:
 
 .. Tell us a bit about yourself.
 
-FILLME
+I am srinivas , an 18 year old second year undergraduate student from Keshav Memorital Institute of Technology ,Hyderabad, India.
+I am a tech enthusiast and a person who loves sports.
 
 Background Information
 ----------------------
@@ -102,31 +103,47 @@ Background Information
 similar programmes before?  If so, tell us about how it went, and any areas you
 would have liked more help with.**
 
-FILLME
+No
 
 **Please tell us about any previous experience you have with Xapian, or other
 systems for indexed text search.**
 
-FILLME
+I dont not have any previous experience with xapian or any other indexed text search.
 
 **Tell us about any previous experience with Free Software and Open Source
 other than Xapian.**
 
-FILLME
+There is a software Barrier,an Open source KVM Software where you can use one keyboard and mouse 
+on different computers nearby when all are connected to same network..I was trying their source code 
+and have implemented few small c programs in linux for capturing device events from one computer and 
+sending to other system in the local network with known IP address.
 
 **What other relevant prior experience do you have (courses taken at college,
 hobbies, holiday jobs, etc)?**
 
-FILLME
+I have taken course on Machine Learning and its science applications where I have learnt the basics of machine learning 
+and deep learning and have worked with Convolutional Neural Networks and facial recongnition,
+a course on C programming and assembly language where I have learnt how C code gets compiled and 
+to use inline assembly in C for 8086 microcontroller, I am familiar with cuda and parallel programming. 
+I have worked with small scale electronics  and IoT projects(with raspberry,arduino).
+I love problem solving and I keep learning different algorithms and datastructures.
+My hobbies are playing cricket and exploring new places. 
+
+
 
 **What development platforms, tools and methods do you prefer to use?**
 
-FILLME
+I prefer Visual Studio Code when I'm learning a new language becuase it provides auto-completion and gives information like 
+what that particular class or function is about, when I get the basics I switch 
+to vim ,the editor I use for more focus , speed and accuracy. 
+I use Docker for testing in linux and recently started using Travis CI . 
+I prefer Guake terminal for fast switching between code and terminal.
+
 
 **Have you previously worked on a project of a similar scope?  If so, tell us
 about it.**
 
-FILLME
+No 
 
 **What timezone will you be in during the coding period?**
 
@@ -134,7 +151,7 @@ FILLME
 .. name so we aren't surprised by any differences around daylight savings
 .. time, which don't all line up in different parts of the world.
 
-FILLME
+GMT+5.30 (IST) 
 
 **Will your Summer of Code project be the main focus of your time during the
 program?**
@@ -143,7 +160,7 @@ program?**
 .. but if we don't know about them in advance we can't make sure you have
 .. the support you need.
 
-FILLME
+Yes , Summer of Code is my main focus this summer.
 
 **Expected work hours (e.g. Monday–Friday 9am–5pm UTC)**
 
@@ -162,7 +179,7 @@ FILLME
 .. these are few, it is usually possible to get enough done across Summer of
 .. Code to make for a worthwhile project.
 
-FILLME
+Monday–Saturday 5am-4pm UTC 
 
 **Are you applying for other projects in GSoC this year?  If so, with which
 organisation(s)?**
@@ -171,24 +188,27 @@ organisation(s)?**
 .. we don't have a problem with that, but it's helpful if we're aware of it
 .. so that we know how many backup choices we might need.
 
-FILLME
+No
 
-Your Project
-============
+Implement Bindings for Go
+==========================
 
 Motivations
 -----------
 
 **Why have you chosen this particular project?**
 
-FILLME
+My main areas of interest includes C++ and Go. They provide greater flexibility and concurrency.
+I have previously  used cgo and swig in small projects.
+So I feel I am elgible for this project.
 
 **Who will benefit from your project and in what ways?**
 
 .. For example, think about the likely user-base, what they currently have to
 .. do and how your project will improve things for them.
 
-FILLME
+As Go lang is emerging now a days rapidly and its being used in servers and distributed systems,
+providing a Go API for the people who want to use xapian search engine in thier servers would be a huge benefit. 
 
 Project Details
 ---------------
@@ -197,27 +217,32 @@ Project Details
 
 **Describe any existing work and concepts on which your project is based.**
 
-FILLME
+I have done some intial implementation of go bindings for xapian which are available in my git repository (currently working on linux and unix systems).
 
 **Do you have any preliminary findings or results which suggest that your
 approach is possible and likely to succeed?**
 
-FILLME
+Yes, my intial work on bindings definitly show that bindings can be implementend in go with c++ shared libraries.
+For time being they are working on Linux and Unix systems. I consider expanding them to work on other platforms is not a difficult job. 
 
 **What other approaches have you considered, and why did you reject those in
 favour of your chosen approach?**
 
-FILLME
+Initially I was able to provide bindings after installation of the shared library on the host system
+, but then I have realised that before installing the core we need to build the bindings and I was using 
+pkg-config , later I started using flags from configure as one of my mentor (olly) suggested.
 
 **Please note any uncertainties or aspects which depend on further research or
 investigation.**
 
-FILLME
+I need to have a closer look at iterator classes of xapian for Implementing bindings in go.
 
 **How useful will your results be when not everything works out exactly as
 planned?**
 
-FILLME
+I am confident enough that the project would be implemented and completed properly on time.
+If in case some thing goes wrong , I would continue my work after GSoC timeline as I am very much 
+interested in this project and love to contribute to xapian. 
 
 Project Timeline
 ----------------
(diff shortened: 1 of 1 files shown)


proposal.rst
```diff
@@ -67,27 +67,28 @@
 .. https://teom.wordpress.com/2012/03/01/how-to-write-a-kick-ass-proposal-for-google-summer-of-code/
 
 ======================================
-FILLME WITH THE TITLE OF YOUR PROPOSAL
+GSOC 2020 - PROPOSAL
```
jaylett: The title of your proposal should go here.

proposal.rst
```diff
@@ -162,7 +179,7 @@ FILLME
 .. these are few, it is usually possible to get enough done across Summer of
 .. Code to make for a worthwhile project.
 
-FILLME
+Monday–Saturday 5am-4pm UTC
```
jaylett: I'd suggest that 11 hours a day six days a week is not really sustainable for the entire length of the programme. While this is obviously up to you, please take seriously our guidance notes in the comment above — because if you burn out in the first few weeks by pushing yourself too hard, that will almost certainly result in you being failed at one of the later checkpoints (because you'll be unable to contribute effectively after you've burned out). You can generally get as much or more done sustainably (so over an indefinite period) by keeping your regular hours down to more like 40 hours a week. While we don't want to discourage your enthusiasm, we see this as a valuable part of the programme: learning how to contribute in a sustainable fashion, whether to open source or to other kinds of jobs.

proposal.rst
```diff
@@ -171,24 +188,27 @@ organisation(s)?**
 .. we don't have a problem with that, but it's helpful if we're aware of it
 .. so that we know how many backup choices we might need.
 
-FILLME
+No
 
-Your Project
-============
+Implement Bindings for Go
```
jaylett, line 193: We meant you to put the project title at the top of the document rather than here. It's not a huge problem though!

proposal.rst
```diff
@@ -171,24 +188,27 @@ organisation(s)?**
...
+==========================
 
 Motivations
 -----------
 
 **Why have you chosen this particular project?**
 
-FILLME
+My main areas of interest includes C++ and Go. They provide greater flexibility and concurrency.
```
jaylett: It isn't hugely important, but if you make a statement like "X provides greater Y", as you do here, this is meaningless without something to compare _to_. "X provides greater Y than Z" makes sense. It's almost never true that X provides greater Y than _everything else it could be compared to_, simply because of the enormous range of technologies out there. (Also, "concurrency" is a tricky beast. Do you mean concurrency within a single process, using threads and fibers; concurrency within a single machine or container which might also use multiple processes; concurrency within multiple machines in a cluster located closely within the network and/oir hardware topology; concurrency across arbitrary machines on the internet; and so on. I'm sure you know what you mean here, but it's always worth taking time to step back and think about whether you've provided enough detail to communicate that.)

A specific reason I'm pointing this out is that "<my favourite thing> provides greater <something or another>" is commonly seen online in people who get very close to one particular set of technologies, and believe they're better than everything else — but don't stay sufficiently up to date with other technologies to be able to know that for sure. So these kinds of statements online tend to be disregarded as being more about the person making them than about technology. It's good practice therefore to avoid statements like this, and instead be more precise about what you mean.

proposal.rst
```diff
@@ -267,8 +292,69 @@ Project Timeline
...
+
+Before May 18,2020 :
+I would finish up the build system part of work for go bindings that 
+are generated purely from xapian-headers.i and xapian-head.i such that
+these can be used like any other bindings and in any platform that other xapian bindings can be used in.
+
+Each week for first 3 days I would work on provding API , that is writing swig interface 
+for the functions and classes which aren't generated from 
+existing swig interfaces(xapian-headers.i and xapian-head.i) with typemaps and other features of swig in go.i.
+Next 2 days documenting the code with examples and the last day for testing.
```
jaylett: When you talk about testing here, do you mean automated tests?

proposal.rst
```diff
@@ -267,8 +292,69 @@ Project Timeline
...
+
+    3.  Provide the go get feature for using xapian bindings in go with 
+        proper guide lines regarding whether the xapian core is installed
+        or not.
+
+
+Before May 18,2020 :
+I would finish up the build system part of work for go bindings that 
+are generated purely from xapian-headers.i and xapian-head.i such that
+these can be used like any other bindings and in any platform that other xapian bindings can be used in.
```
jaylett: A word of warning: unless you have access to a Windows machine (or VM) it may be difficult to get this working. There's a good chance that the go build system will have wrinkles specific Windows. What I'd suggest is that you move that to a stretch goal to the overall project. (Fully-functional go bindings that work on linux alone is better than partially-functional bindings that work on linux, Windows & macos.)

proposal.rst
```diff
@@ -267,8 +292,69 @@ Project Timeline
...
+for the functions and classes which aren't generated from 
+existing swig interfaces(xapian-headers.i and xapian-head.i) with typemaps and other features of swig in go.i.
+Next 2 days documenting the code with examples and the last day for testing.
+I have grouped different classes for each week.
+
+
+First Month : 
+
+  Week1 :
+        Work on Document , Registry , Query , Stem implementation , Stem classes.
```
jaylett: You've listed a bunch of classes in each week, but it's not clear what you intend to do for them. Generally there are going to be several things you need to do for each class:

1. Wrapping anything that SWIG doesn't do automatically. This may involve typemaps, or may involve more complex custom code and/or additional SWIG directives.
2. Making the resultant API more idiomatic for golang users. Typically the main wins are in iterators, but there may be other things that better match go if they are abstracted further away from the C++ API.
3. Making sure exceptions work as intended. golang doesn't have exceptions; instead it has a return code pattern.
4. Supporting Directors for the class if necessary (which is a SWIG concept allowing you to subclass a C++ class in the target language). This only applies to some classes; most of Xapian you don't need or want to subclass. (Document, for instance, isn't helpful to subclass. MatchSpy is.)

What I think you should do as much as possible this week is to look through the different classes you've listed, and figure out which of the things above make sense to do. Then you may want to think in terms of functionality to deliver piece by piece. For instance, you might want to first get to a place where you can write a simple query system in golang is probably helpful. For that, you need at least Database, Document, Enquire, Query, QueryParser, MSet and MSetIterator, and probably Stem (but you're unlikely to need StemImplementation) — possibly with some methods ignored. Then you could add basic indexing (WritableDatabase and TermGenerator), and then gradually fill out the other functionality (which will depend on what you think is most useful — perhaps you want to add support for Stopper, or for the other iterators, or for configuring the weighting scheme used when querying).

A good way of figuring out the chunks of functionality would be to run through the getting started / user guide, which introduces different classes and functionality in a sensible order.

proposal.rst
```diff
@@ -267,8 +292,69 @@ Project Timeline
...
+  Week7 :
+        Work on all the Derived classes of Posting Source.
+  Week8 :
+        Work on all the Derived classes of Matchspy classes and classes related to Latlong.
+
+Third Month : 
+  Week9 :
+        Work on Database class.
+  Week10 :
+        Work on Exceptions.
```
jaylett: The work in these weeks (9 & 10) are almost certainly more important than PostingSource, MatchSpy, and some of the others you have in the second month. However you might be able to split the work here — ensure enough of the Database functionality is working for most uses, then move on to things like MatchSpy before coming back to tidy up the rest of Database. The more you can get into this level of detail, the easier it will be to make a success of the project — so we expect quite a lot of detail in the proposal.

jaylett (changes requested):
The main thing I'd suggest is to get more into the detail of what you need to do. When you're talking about working on a class like Document, what will you need to do? What does SWIG already give you for free? Will you be adding new golang code, new C++ code, new typemaps, or something else?

It's probably worth looking at this project description and plan: https://trac.xapian.org/wiki/GSoC2011/SupportLua/ProjectPlan

It was from a proposal a number of years ago for adding Lua support using SWIG. It was a successful project, although I'd have preferred to see more detail in the timeline / project plan at the end, in the way I've indicated in some of my comments here.

ojwb:
> It was from a proposal a number of years ago for adding Lua support using SWIG. It was a successful project, although I'd have preferred to see more detail in the timeline / project plan at the end, in the way I've indicated in some of my comments here.

It's probably worth noting that Xiaona Han had already managed to get basic Lua bindings working from scratch before we were reviewing proposals, so we had a lot of confidence they could make a success of the project.

proposal.rst
ojwb: It would certainly be good to get them working portably, but I'd agree that getting them working well on one platform first is probably a better approach.  Windows and macOS both have a number of differences from Unix-like platforms that can suck up a lot of time.

proposal.rst
ojwb: There's probably nothing needed over the default SWIG wrapping for some classes.  For example, there's no special handling for `Xapian::Registry` in any of the existing bindings, so I doubt any would be needed for Go (try `git grep -w Registry` in xapian-bindings to see this).

proposal.rst
```diff
@@ -351,8 +352,8 @@ Plan :
         
         1. Using channels one could use for-range construct.
 
-        for i := range container.Iter(){
-          i.GetData() // methods to get information from the iterator at that position.
+        for i := range doc.terms(){
+          i.GetTerm() // methods to get term from the iterator at that position.
```
ojwb, line 389: This sounds like a good way to wrap these, but what's actually involved in doing so?

srinivasyadav18: you mean to show me how its done internally in go?...If yes then ill code it and show...thats done using channels in go ... in a brief range gets data from the channel which is open until iterator is reaches the last element , then channel closes and range stops..

srinivasyadav18: > This sounds like a good way to wrap these, but what's actually involved in doing so?

Hey I have added a link how this can be done with small piece of code.

ojwb: Kind of, but that's just showing how to create a channel using only Go code.  The real (and perhaps more tricky, though perhaps not) question is what's actually involved in doing the same but with data coming from a Xapian iterator wrapped via SWIG.

srinivasyadav18: > Kind of, but that's just showing how to create a channel using only Go code. The real (and perhaps more tricky, though perhaps not) question is what's actually involved in doing the same but with data coming from a Xapian iterator wrapped via SWIG.

Hey I have put the code for implementing termiterator in document class (I added only raw code how termiterator is implemented but the other parts of document interface needs rewrapping which will be done week 2 , so please consider only for-range construct  implementation part)

srinivasyadav18 (commented):
Thank You for the feedback. I have made the changes to proposal please review it.

proposal.rst
```diff
@@ -267,8 +304,191 @@ Project Timeline
...
+        2. Using methods such as Iter.Next() as used in Go lang standard library (Container List https://golang.org/pkg/container/list/).
+         
+        Both standard method /* for iter.Next(){ ... code } */ and /* for-range construct would be made available for user
+        (these are just like auto-ranged based loops in c++) */
+         
+        Go support multiple return values , therefore rewrapping the interfaces which return iterators to both 
+        begin and end iterators in one function call as below.
+        /* begin,end := doc.Terms() */ 
+
+      * Go supports errors as return values . A language like c++ have try catch block Go has three constructs for dealing
```
jaylett, line 460: You talk about handling exceptions using panic/defer/recover, but is this the right approach? Should our exceptions instead be converted into return values? Can you write some example golang showing how to handle a common case? (The most obvious is probably when the `Database()` raises an exception if you tell it to open a database that doesn't exist.)

srinivasyadav18: > You talk about handling exceptions using panic/defer/recover, but is this the right approach? Should our exceptions instead be converted into return values? Can you write some example golang showing how to handle a common case? (The most obvious is probably when the `Database()` raises an exception if you tell it to open a database that doesn't exist.)

Hey I have put the example case for Database already...please check, it catches the error as return value during call of the constructer. I have followed the pattern that go  standard library follows.

proposal.rst
srinivasyadav18: > Kind of, but that's just showing how to create a channel using only Go code. The real (and perhaps more tricky, though perhaps not) question is what's actually involved in doing the same but with data coming from a Xapian iterator wrapped via SWIG.

Hey , I have put the code for iterators , errors , type conversions , rewrapping . Please review it. Thank You 😊

ojwb: > https://github.com/srinivasyadav18/xapian-gsoc-plan/blob/master/go.i#L43

This looks promising, though it seems only provides access to the term name.

What about the other accessor methods on the `TermIterator` object?  At least for some of them, they can involve significant work so just reading all of them up front isn't great.  The Python bindings have attempted to address such issues already - maybe that's worth looking at to see if the approach works for Go too.

It's also good to try to partly automate such wrapping via SWIG macros - otherwise you're going to end with a lot of code cut-and-pasted and then modified to support all the different methods which return a `TermIterator`, and that's not great for maintainability.

proposal.rst
```diff
@@ -369,6 +369,9 @@ Plan :
         with exceptions, they are panic defer and recover.A Panic is similar to an exception which can occur an runtime exception.
         C++ exceptions can be handled in go from swig wrappers as follows(https://github.com/srinivasyadav18/xapian-gsoc-plan/blob/master/example.i#L16).
         Which ever class function throws an exception in c++ , the wrapped function in Go returns the error as value.
+        Database error handling - https://github.com/srinivasyadav18/xapian-gsoc-plan/blob/master/go.i#L77
```
ojwb, line 464: We really need to find a way to automate such error handling - having to write this by hand for every wrapped method will make the Go bindings too hard to maintain going forwards.

For a typical new C++ method, we get automatic wrapping for all the bindings for **zero** extra work.  In some cases, extra work to get a new C++ method wrapped is needed, but it's rare.

Having to do something for every new wrapped method scales badly in terms of developer effort, and we definitely need to avoid that.  We've had non-SWIG-based bindings which required manual update before (XS bindings for Perl and JNI bindings for Java) and they constantly lagged behind the C++ API, and that frustrated both users and developers.

proposal.rst
```diff
@@ -369,6 +369,9 @@ Plan :
...
+        https://github.com/srinivasyadav18/xapian-gsoc-plan/blob/master/main.go#L36
+        error (pastebin) - https://pastebin.com/AuFpiRdQ
```
ojwb, line 466: Please include content in your proposal rather than linking to external sites.  The final proposal you submit is essentially what you're saying you'll do if chosen, so it ought to be self-contained rather referring out to information on external sites can change or even disappear (content on paste sites typically expires after a while so is particular poor to refer to).

It would also be better to inline or include the content referred to in https://github.com/srinivasyadav18/xapian-gsoc-plan/blob/master/main.go for the same reasons.  That's at least that's under version control, so not quite as bad as linking to a paste site.

proposal.rst
srinivasyadav18: > We really need to find a way to automate such error handling - having to write this by hand for every wrapped method will make the Go bindings too hard to maintain going forwards.
> 
> For a typical new C++ method, we get automatic wrapping for all the bindings for **zero** extra work. In some cases, extra work to get a new C++ method wrapped is needed, but it's rare.
> 
> Having to do something for every new wrapped method scales badly in terms of developer effort, and we definitely need to avoid that. We've had non-SWIG-based bindings which required manual update before (XS bindings for Perl and JNI bindings for Java) and they constantly lagged behind the C++ API, and that frustrated both users and developers.

Since errors are treated as values in Golang I am explicitly hand wrapping for the function that throws exception. I would work on it and find a way to automate for wrapping error  in community bonding period.

proposal.rst
srinivasyadav18: > Please include content in your proposal rather than linking to external sites. The final proposal you submit is essentially what you're saying you'll do if chosen, so it ought to be self-contained rather referring out to information on external sites can change or even disappear (content on paste sites typically expires after a while so is particular poor to refer to).
> 
> It would also be better to inline or include the content referred to in https://github.com/srinivasyadav18/xapian-gsoc-plan/blob/master/main.go for the same reasons. That's at least that's under version control, so not quite as bad as linking to a paste site.

Is it fine the proposal will get longer? I have include the code now in proposal . If it is not good please give a feedback. I have put even the links at the end in references.

proposal.rst
srinivasyadav18: > We really need to find a way to automate such error handling - having to write this by hand for every wrapped method will make the Go bindings too hard to maintain going forwards.
> 
> For a typical new C++ method, we get automatic wrapping for all the bindings for **zero** extra work. In some cases, extra work to get a new C++ method wrapped is needed, but it's rare.
> 
> Having to do something for every new wrapped method scales badly in terms of developer effort, and we definitely need to avoid that. We've had non-SWIG-based bindings which required manual update before (XS bindings for Perl and JNI bindings for Java) and they constantly lagged behind the C++ API, and that frustrated both users and developers.

Hey do you mean to say if a new c++ method gets added then that should get rewrapped automatically instead of hand wrapping?if that's the case..I think most of the work is in writing a script which generates the wrappers automatically. You are correct, we get automatic wrapping for other bindings with no extra work, but the swig support for Go doesn't produce the API which a go user can use naturally
. So this means that we need to provide second level of wrapper (kind of another swig software ) to generate wrappers for swig generated objects to produce a simple API.

proposal.rst
srinivasyadav18: Hey that’s not a big issue ..I could just send the term iterator object over the channel ..so now through iteration we can access all the method of it.

proposal.rst
srinivasyadav18: Could you please explain the comment in detail..I didn't understand completely

ojwb: > I think most of the work is in writing a script which generates the wrappers automatically

We already have a tool which generates wrappers automatically - it's called "SWIG".

You should be using SWIG to achieve this, not writing extra layers of scripts above or around it.  SWIG is up for the job of automating the majority of the wrapping for all the other languages we use it for - I don't see why Go is different in that regard.

proposal.rst
ojwb: Don't worry about length (at least if the content is relevant as this is).

ojwb:
Note you'll need to submit a "final" PDF version of this at https://summerofcode.withgoogle.com/ to be eligible for GSoC - currently I only see a draft version there, and Google require a final version before the deadline or else we can't select you.

You should be able to generate a suitable pdf from the .rst version using make pdf.

ojwb:
Thanks for you proposal.

PR #18, opened 2020-03-23
Evaluation Suite

slesarev-hub:


proposal.rst
```diff
@@ -162,7 +161,7 @@ FILLME
 .. these are few, it is usually possible to get enough done across Summer of
 .. Code to make for a worthwhile project.
 
-FILLME
+Monday–Friday 7am-1am UTC, but I can move it if needed.
```
jaylett: You seem to be suggesting you'll work 18 hours a day, for a total of 90 hours a week. I'd gently suggest that this is far too much to contribute at your best capability for a sustained project across months.

proposal.rst
```diff
@@ -268,7 +267,29 @@ Project Timeline
...
+
+In each next position assumes that testing and writing cross-platform part is included.
+
+* 1 Week - integrate datasets 
+
+* 2 Weeks of exams (situation with exams is very unstable now in my University, so I timeline may change in this case)
+
+* 1 Week - solve possible problems with the size of datasets
+
+* 3 Weeks - implement ListMLE, ListNet and SVM evaluation api
```
jaylett: We're expecting quite a bit more detail than this, particularly for a 3 week segment. What do you mean by an "evaluation API" in this case? How will it relate to the previous work on xapian-evaluation mentioned in the letor stabilisation project page (linked from the evaluation suite project description)? How will it related to the `Scorer` classes already in the letor codebase? What would be different about the support for ListMLE, ListNet and SVM? Are you going to tackle the outstanding PRs from last year's letor project, or is this just about evaluation (in which case why are you focussing it on the letor rankers, rather than all weighting approaches)?

It's probably worth taking a step back and explaining in your proposal how and why someone would use this. You say in your motivation answer that it will be "useful for users [who] want to understand Xapian abilities", but ideally an evaluation system would also allow them to choose which ranking algorithm (including letor Ranker) and parameters works best for their case — and to allow the Xapian developers to keep track of which ones work best for which common real-world examples, so we can choose appropriate defaults that work well in general use.

I think it'd be good to sketch out how someone would use this project. Is it a tool that people run? (A bit like xapian-evaluation.) Is it a set of classes in Xapian? (A bit like Scorer.) Or perhaps it's a tool built on internal classes, so people can use the metrics in their own tools, or can evaluate performance using an easy-to-run one.

Also, what datasets are you going to evaluate over? It seems natural that part of this project would involve running the evaluation suite, for which you'll need a corpus and some test queries (ideally pre-labelled). A problem that vivekp ran into was that different datasets have different formats, so either you need to have a single format that is documented that others can be transformed into, or maybe you'll need a library of support for different formats.

proposal.rst
```diff
@@ -268,7 +267,29 @@ Project Timeline
...
+
+If no troubles appeared, I can try to develop suite for performance testing:
+
+* 1 Week - get familiar with data indexing and searching api
+
+* 1.5 Weeks - tests for indexing
+
+* 1.5 Weeks - tests for searching
```
jaylett: Again, we expect to see more detail. What are you actually aiming to test here? The project description mentions time taken, memory used, and disk space used — but there are other options (such as user + system CPU time) that could be useful as well.

As with the evaluation side, what dataset(s) do you want to work with?

proposal.rst
```diff
@@ -313,4 +334,4 @@ you plan to use, please give details.**
 .. to clearly identify that code (and keep existing licensing and copyright
 .. details intact), and to check with the mentors that it is OK to use.
 
-FILLME
+I don't know exactly, maybe some third-party api for interacting with big datasets will be needed.
```
jaylett: What problems do you think you might encounter that would lead to needing this? Where would they fit within Xapian (which itself is already a system for dealing with at least medium-sized datasets, depending on your definition of size)?

proposal.rst
```diff
@@ -267,29 +285,29 @@ Project Timeline
...
 .. in your project timeline.
 
-* Weeks of bonding: get familiar with ranking algorithms, search for suitable datasets, assessment of the complexity of integrating large databases, and performance tests.
 
-In each next position assumes that testing and writing cross-platform part is included.
+* Weeks of bonding: Get more familiar with ranking algorithms, select specific suitable parts of the above datasets, discuss what is wrong with `#278 <https://github.com/xapian/xapian/pull/278>`_.
 
-* 1 Week - integrate datasets 
+* 1 Week - FIRE dataset integration. Each user have to download it by themself, because of 
+registration required. Implement training process using Letor API.
```
jaylett, line 304: What will it mean to implement the training process? Do we have a parser for FIRE, or one close enough to adapt, or will you be writing something from scratch? What does "implement training process" mean here? The letor API provides all the hooks for doing this, so is this adding more functionality to xapian-evaluation, or some new tool?

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
 
 Project Details
 ---------------
 
-.. Please go into plenty of detail in this section.
-
 **Describe any existing work and concepts on which your project is based.**
 
-There are several ranking algorithms with pretty same api implemented in Letor. So I think manager class should be developed to evaluate ranking and logging results. It will better to integrate it with testing classes, because it makes possible to develop suite for perfomance rating, wich will useful not only for ranking, but for all test cases. 
+There are several ranking algorithms with pretty same api implemented in Letor. So I think manager class should be developed to evaluate ranking and logging results. It will better to integrate it with Letor classes. Additional task is to improve perftest suite by running it on real data, wich will useful not only for ranking benchmark, but for all test cases. More detailed:
```
jaylett: I think you should try to spell out in more detail what you mean by a manager class to evaluate ranking and logging results. That sounds like two different pieces of work, so it's not immediately clear why one manager would deal with this.

Evaluation (how well we rank documents for queries) and performance testing are very different things. If you want to tackle them both, then I'd treat them as totally different subprojects, which means they need their own rationale, their own description and project breakdown, and their own steps in the timeline.

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
+
+* Implement class for preparing data, i.e. download it and convert to suitable form. The most convenient dataset is `FIRE <http://fire.irsi.res.in/fire/static/data>`_ and `TREC <https://trec.nist.gov/data/t2002_filtering.html>`_. It is hard to fully generalize this case, so I prefer to implement parser for each dataset:
+
+	* FIRE queries stored in inappropriate format, they should be transformed to use them in Xapian; It will possible to use English language corpus, because Eglish stemmer exists in Xapian.
+
+	* TREC dataset has big enough amount of documents, it is possible to combine some of them during pasing to create more complex queries.
+
+* Implement evaluator. According to xapian-letor/tests/api_letor.cc it is no big difference in applying rankers to data, it should just apply algorithms and return scores, so it could be another one class in xapian/letor. It would be nice to be able to call it same way as make check.
```
jaylett: How can you make this as simple as `make check`? Will there be a standard place to store the source corpus, queries, and qrels? What do you actually want people to be able to run? Writing out the steps that someone will have to take to use this will help you break down what you need to implement, as well as ensure we understand clearly what you are proposing.

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
+
+* Write suite for combining documents. TREC dataset has big corpus of small documents. To get more content small documents could be merged. Also queries and qrels have to be transformed. After that evaluation could be done using Xapian API logic operations for queries. Since it is not needed for every Letor run, it should be separate tool, like xapian-letor/bin/xapian-prepare-trainingfile.cc.
```
jaylett: If you combine documents, are the qrels going to make any kind of sense? Is there some paper or technique you're proposing to transform the qrels? Also, since you subsequently talk about wikipedia (which has large as well as small documents), do you actually need to do this?

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
+There are several ranking algorithms with pretty same api implemented in Letor. So I think manager class should be developed to evaluate ranking and logging results. It will better to integrate it with Letor classes. Additional task is to improve perftest suite by running it on real data, wich will useful not only for ranking benchmark, but for all test cases. More detailed:
+
+* Implement class for preparing data, i.e. download it and convert to suitable form. The most convenient dataset is `FIRE <http://fire.irsi.res.in/fire/static/data>`_ and `TREC <https://trec.nist.gov/data/t2002_filtering.html>`_. It is hard to fully generalize this case, so I prefer to implement parser for each dataset:
+
+	* FIRE queries stored in inappropriate format, they should be transformed to use them in Xapian; It will possible to use English language corpus, because Eglish stemmer exists in Xapian.
```
jaylett: What is inappropriate about the format in FIRE? What are you going to transform them to?

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
+
+* Implement evaluator. According to xapian-letor/tests/api_letor.cc it is no big difference in applying rankers to data, it should just apply algorithms and return scores, so it could be another one class in xapian/letor. It would be nice to be able to call it same way as make check.
+
+* Write suite for combining documents. TREC dataset has big corpus of small documents. To get more content small documents could be merged. Also queries and qrels have to be transformed. After that evaluation could be done using Xapian API logic operations for queries. Since it is not needed for every Letor run, it should be separate tool, like xapian-letor/bin/xapian-prepare-trainingfile.cc. 
+
+* It is possible to use Wikipedia dumps as data for perftest. `Wikipedia <https://dumps.wikimedia.org/enwiki/latest/>`_ corpus just stores XML file with links to articles and some info, so it each document should be downloaded and indexed individually. Also plain text `extractor <https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/corpora/wikicorpus.py>`_ and query fabricator needed.
```
jaylett: Why should each document be downloaded individually? Can't you index from the single wikidump?

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
 
 **Do you have any preliminary findings or results which suggest that your
 approach is possible and likely to succeed?**
 
-I worked with Xapian test harness, it has BackendManager class wich makes testing databases api easier. I believe ranking benchmark should be done in similar way. 
+I worked with Xapian test harness in `#758 <https://trac.xapian.org/ticket/758>`_, it has BackendManager class wich makes testing databases api easier. I believe ranking benchmark should be done in similar way.
```
jaylett: This isn't really a preliminary finding; it's a belief you have based on looking at something quite different. You should indicate in detail how you want the ranking benchmark to be used by people.

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
 
 **What other approaches have you considered, and why did you reject those in
 favour of your chosen approach?**
 
-Another way is to create ranking suite as part of Letor, but it is not clear enough how to develop performance measurement api in this case.
+I have discussed the possibility of using web datasets like INEX, but there are several reasons to store dataset locally:
```
jaylett: INEX has at least some downloadable datasets, so I don't understand what problem you're articulating.

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
+
+* Most of Web-API datasets requires an account creation, which is not suitable for Xapian users.
+
+* All data should be stored locally at least in parts to be indexed.
```
jaylett, line 235: Is this true? It might be more convenient, but "should" is a strong word. If you're indexing a million 1MB documents, that's a lot of storage that you'll need: for the source documents, then more for the database you're building. In that situation, would it be more useful to be able to index online documents instead?

(I'm *not* saying that you should take this route. I'm saying that you should be more precise in your statements in the proposal. If you say something like "we should do things this way" then you need to back that up with either references to existing work or research, or your own argument.)

proposal.rst
```diff
@@ -180,39 +180,57 @@ Motivations
...
 
 **Please note any uncertainties or aspects which depend on further research or
 investigation.**
 
-Databases could be too big to store them on local machine, or too big for memory, which could cause inaccuracies in performance tests, because of reading from disk. Also I don't know how different ranking algorithms are, so it could impact timeline.
+Databases indexing time could be big enouth, in connection with this, difficulties may arise during implementation.
```
jaylett: Big enough for what? What difficulties are you worried about? (There are always difficulties in implementation, but if you have specific ones you're concerned about, make that clear.)

proposal.rst
```diff
@@ -267,29 +285,29 @@ Project Timeline
...
 
-* 1 Week - integrate datasets 
+* 1 Week - FIRE dataset integration. Each user have to download it by themself, because of 
+registration required. Implement training process using Letor API.  
 
-* 2 Weeks of exams (situation with exams is very unstable now in my University, so I timeline may change in this case)
+* 2 Weeks of exams (situation with exams is very unstable now in my University, so I timeline may change in this case).
 
-* 1 Week - solve possible problems with the size of datasets
+* 1 Week - add function for selecting scorer and ranker. It would be nice to write API allows user to select default sets of queries to run and run specified queries. For that purpose update build system to be able to run suite from terminal.
```
jaylett: The evaluation suite can currently be run from the terminal, so I'm not sure what you're going to change here.

When you say "add function for selecting scorer and ranker", are you thinking something like the Registry we use for things like Weight classes, or something like `Weight::create()`? Most people writing code would just instantiate the classes they need, although in the evaluation suite there is a need for people to set configuration (somehow — and you should be clear how you want them to do this) saying which ranker(s) to evaluate.

Do you need people to select different scorers? Is there a reason not to run all of them and report the results?

proposal.rst
```diff
@@ -267,29 +285,29 @@ Project Timeline
...
+* 1 Week - TREC dataset integration. It has similar format to FIRE, it should not be problem to integrate it. Start writing tool for combine docs.
 
-If no troubles appeared, I can try to develop suite for performance testing:
+* 1 Week - combine tool API should support plain logical operations, it could take time to implement them.
 
-* 1 Week - get familiar with data indexing and searching api
+* 1 Week - prepare tool for text data extracting from Wikipedia.
 
-* 1.5 Weeks - tests for indexing
+* 1 Week - add functionallity to perftest to drop system cache for more accurate results and run one case multiple times to get average time consuming.
```
jaylett: What do you mean by "system cache"? Are you talking about the VM / filesystem cache that most Unix-like operating systems have? How are you going to drop that? Is a run with a cold cache indicative of real-life use? Are you talking about running multiple times dropping the cache each time, or dropping the cache at the beginning then running a number of times? How many times is sensible? Should that be user configured?

proposal.rst
```diff
@@ -267,29 +285,29 @@ Project Timeline
...
 
-* 1.5 Weeks - tests for searching
+* 1 Week - implement Wikipedia data loader and queries for it, write performance tests for filling db with documents, iterating over documents.
 
-* 1 Week - solving problems and integrate tests to build system
+* 1 Week - add perftest to autotools srcipt, as it advised in `#107 <https://trac.xapian.org/ticket/107>`_. If time is left, add function to fork test case in new process and measure it memory usage.
```
jaylett: Why do you need to fork to measure memory usage? Can you not use `getrusage()` or similar within the performance test? Or are you worried about significant overhead from the performance test itself? Are you going to check memory usage several times and graph it? How are you going to report what you find?

jaylett (commented):
You've started to add more detail, but I think you can dig a lot deeper. I've tried to give you some suggestions, both in this review and my previous one, of questions you can ask yourself to get into the details of what your project will entail. The more you plan up front, the more you understand before you start writing code. That generally allows you to do more in the same amount of time.

proposal.rst
ojwb: And how do you decide which groups of documents to combine?

Just grouping documents arbitrarily (e.g the first 10 docs together, then the next 10 together, etc) would produce larger documents, but they won't be representative of what a real dataset with 10x larger documents would be like.

Even combining small documents on the same subject may not produce a realistic larger document.  Think what the concatenated text of 1000 tweets on the same subject would look like - nobody would write a (say) 30,000 word document which looked like that.

Unless you have a good technique for this, it'll be better to find a test dataset with a larger document size.

proposal.rst
ojwb: I think @slesarev-hub probably looked at the wrong file.  There are various different dumps with different contents, but `enwiki-latest-pages-articles.xml.bz2` (or similar) contains the full text of all the articles.

We definitely shouldn't be automated the fetching of individual articles - such mass downloading is abusive of wikipedia's servers and liable to get people blocked.

proposal.rst
```diff
@@ -268,7 +285,29 @@ Project Timeline
...
+
+* 1 Week - combine tool API should support plain logical operations, it could take time to implement them.
+
+* 1 Week - prepare tool for text data extracting from Wikipedia.
+
+* 1 Week - add functionallity to perftest to drop system cache for more accurate results and run one case multiple times to get average time consuming.
+
+* 1 Week - implement Wikipedia data loader and queries for it, write performance tests for filling db with documents, iterating over documents.
+
+* 1 Week - add perftest to autotools srcipt, as it advised in `#107 <https://trac.xapian.org/ticket/107>`_. If time is left, add function to fork test case in new process and measure it memory usage.
```
ojwb: `perftest` is already integrated in the autotools build system - that isn't what #107 is about at all.

The fundamental problem with `perftest` is it is entirely based on randomly-generated data, and that means it isn't representative of real-world use (or at least we can't trust that it is).  That makes profiling and optimisation based on `perftest` results dubious at best.

We want a performance test suite based on real world data, or at least data that's very carefully randomly generated to match real data in ways that matter (but just using real data seems simpler).

ojwb:

ojwb:
Thanks for you proposal.

PR #19, opened 2020-03-25
Weighting Schemes

dipanshu124:
Please Review it .

### DIFF
--- a/proposal.rst
+++ b/proposal.rst
@@ -79,7 +79,7 @@ About You
 
  * IRC nickname(s): dipanshu
 
- * Any personal websites, blogs, social media, etc: FILLME
+ * Any personal websites, blogs, social media, etc: https://www.facebook.com/profile.php?id=100004662888280
 
  * github URL: dipanshu124@github.com
 
@@ -87,7 +87,7 @@ About You
 
 .. Tell us a bit about yourself.
 
-I am a Computer Science student of Indian Institute of Technology ,Kanpur,INDIA.I am proficient in C and C++ .As a hobby ,I practice Competitive Programming . Apart from programming, I am also very fond of Mathematics.
+I am a Computer Science student of Indian Institute of Technology , Kanpur,INDIA.I am proficient in C and C++ .As a hobby ,I practice Competitive Programming . Apart from programming, I am also very fond of Mathematics.
 
 Background Information
 ----------------------
@@ -107,7 +107,7 @@ No,This is the first time.
 **Please tell us about any previous experience you have with Xapian, or other
 systems for indexed text search.**
 
-Before intending to apply for gsoc,I didn't have much experience.Now I have been using Xapian since the past 2 months.I am familiar with its indexed text search and also with the implementation of current Weighting schemes(the part where I intend to work).
+Before intending to apply for gsoc,I didn't have much experience.Now I have been using Xapian since the past 1 month.I am familiar with its indexed text search and also with the implementation of current Weighting schemes(the part where I intend to work).
 
 **Tell us about any previous experience with Free Software and Open Source
 other than Xapian.**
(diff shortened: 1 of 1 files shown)


proposal.rst
```diff
@@ -268,8 +294,64 @@ Project Timeline
...
+ Week 3(18 MAY-24 MAY)
+     Going through Xapian Evaluation files in detail.
+ Week 4(25 MAY-31 MAY)
+     Time for any other discussion needed prior to coding.
+  
+ **Coding**
+  Week 1(1 JUN-6 JUN)
+      - implement Entropy and Global frequency IDF (1 day)
+      - write test cases for Entropy and Global frequency IDF (3 days)
+      - make PR for these changes and getting it reviewed and completing documentation simultaneously (1 day)
```
jaylett: Even though you have some experience with our PR process, I'd suggest leaving more time for the first PR merge of the project. If everything goes more smoothly, you can replan at one of the two "midterm" evaluations.

You talk about documentation here (and in the other subprojects), but it'd be good to understand whether you're talking about API docs, the user guide, or something else. If you want to show how to use some of these new weighting schemes in the user guide, there'll likely be some more work so it might be worth having a subproject to do a number of them in one go.

proposal.rst
```diff
@@ -268,8 +294,64 @@ Project Timeline
...
+      - make changes in trec_search.cc (2 days)
+  Week 10(2 AUG-8 AUG)
+      - write test cases for these changes (3 days)
+      - make PR for these changes and getting it reviewed and completing documentation simultaneously (2 days)
+  Week 11(9 AUG-15 AUG)
+      - Compare effectiveness of different normalisations .
+      - make documentation for these results.
+  Week 12(16 AUG-22 AUG)
+  
+This week is kept as buffer week.Also due to the COVID-19 outbreak in my country,Our exams have been rescheduled but the dates aren't decided.So there is a possibility that I might have exams during the coding period.But they wont last more than a week.
```
jaylett: Are there "stretch" goals you could consider if you get everything done on or ahead of schedule and don't need the buffer?

proposal.rst
```diff
@@ -268,8 +294,64 @@ Project Timeline
...
+      - Add support for the new normalisations in Weight::create().This includes the changes to be made in Registry(3 days)
+      - make PR for these changes and getting it reviewed and completing documentation simultaneously (1 day)
+      - Rest of days are kept free to complete back log if any.
+  Week 9(26 JUN-1 AUG)[Evaluation 2]
+      - submit evaluation.
+      - make changes in config_file.cc (2 days)
+      - make changes in config_file.h (1 day)
+      - make changes in trec_search.cc (2 days)
+  Week 10(2 AUG-8 AUG)
+      - write test cases for these changes (3 days)
```
jaylett: Note that right now xapian-evaluation doesn't have an automated tests at all, so you'll need to figure out what you're actually doing here.

proposal.rst
```diff
@@ -268,8 +294,64 @@ Project Timeline
...
+      - make PR for these changes and getting it reviewed and completing documentation simultaneously (2 days)
+  Week 11(9 AUG-15 AUG)
+      - Compare effectiveness of different normalisations .
```
jaylett: What dataset are you proposing for this?

jaylett (commented):
You have some good detail in here, although since you still have time before submissions close it's always worth seeing if there's more you can add. I would definitely look at adding some stretch goals, since even if you don't have time to get to them it will help you think now about the direction of the project. Something that successful projects often do when they come to their final writeup (the very last step of GSoC) is to propose future ideas that come naturally from the project. Some of that will be stretch goals you didn't get to, although some may come from things you discover during the project. (Some of those proposed future ideas may end up as GSoC projects for someone in a future year!)

proposal.rst
dipanshu124: I have added more time for PR merge as suggested. If even more time is required,I would be happy to utilize the first weekend for first PR merge.
I have updated that I am talking about user guide. In this my aim is to advice users which normalisation they should use depending on their dataset.
If I make the documentation a subproject ,then the time I have in the PR merge process won't be utilized.That is the only reason I am doing documentation in patches.

proposal.rst
dipanshu124: I have now added the Stretch goals.

proposal.rst
dipanshu124: Thanks for pointing this out. I just added this with flow from the test cases of weighting schemes.For now I have removed this.

proposal.rst
dipanshu124: Well I want to use one of MEDLINE, CRANFIELD or CISI dataset.But it seems I dont have access to any of them.
Since Xapian has access to FIRE dataset ,we can use that.

proposal.rst
jaylett: It's important to understand that getting a PR merged — particularly the first one — requires not just time from you, but from us. So you being able to work weekends doesn't help unless we happen to have time then, which isn't always going to happen.

proposal.rst
```diff
@@ -317,48 +317,50 @@ Project Timeline
...
-      - make PR for these changes and getting it reviewed and completing documentation simultaneously (2 days)
   Week 11(9 AUG-15 AUG)
-      - Compare effectiveness of different normalisations .
-      - make documentation for these results.
+      - make PR for these changes (2 days)
+      - Compare effectiveness of different normalisations. We will use FIRE database for this.
   Week 12(16 AUG-22 AUG)
-  This week is kept as buffer week.Also due to the COVID-19 outbreak in my country,Our exams have been rescheduled but the dates aren't decided.So there is a possibility that I might have exams during the coding period.But they wont last more than a week.If I have time,I can work for Stretch goals.
+      - make user guide documentation for these results.
+      - If I have time,I can work for Stretch goals.
```
ojwb: Exams during the coding period are something that happens, and unknown/uncertain dates due to COVID19 restrictions are clearly not under your control, but it's certainly good to note things like this as you had before, so do leave this in unless the uncertainty has since been resolved and the exams are definitely not going to happen during the coding period.

jaylett:
@dipanshu124 if you haven't uploaded your final proposal, you need to do so in the next 40 minutes.

ojwb:
Thanks for you proposal.

PR #20, opened 2020-03-26
GSoC 2020 Project Proposal: Text-Extraction Libraries

Exter-dg:


proposal.rst
```diff
@@ -1,25 +1,26 @@
...
+..
 .. From experience the best proposals are ones that are discussed with us and
 .. improved in response to feedback.  You can share draft applications with
 .. us by forking the git repository containing this file, filling in where
 .. it says "FILLME", committing your changes and pushing them to your fork,
 .. then opening a pull request to request us to review your draft proposal.
+
 .. Please pick a useful title for the pull request - "My application" may
 .. be meaningful to you, but isn't helpful to those reviewing.  Using the
 .. title of the project is a good.  You can open a pull request even before
```
ojwb, line 26: "Draft Proposal" isn't a very useful title - it tells the reader nothing about the proposal, except that it's a draft (and at this stage all proposals are really still drafts).

proposal.rst
```diff
@@ -260,18 +351,53 @@ Project Timeline
...
+
+**Coding Weeks 1-3(June 1 - June 20) Module 3**
+
+  * Research about different zip file reading libraries- eg libarchive,libzip ( Compare them and discuss with the community which is the most appropriate)
+  * Implement code to add support for the library
+  * Test code. Fix Issues if any.
+  * Write proper documentation and publish changes to the project
+
+
+**Coding Weeks 4-11(June 22 - Aug 14) Module 4**
```
ojwb: I think you mean Aug 24 here?

proposal.rst
```diff
@@ -260,18 +351,53 @@ Project Timeline
...
+
+Support for atleast 5 formats will be added during this period. However, I believe I will be able to add more in the given time frame.
+
+A week has been kept free as a buffer in case of any delay in any of the modules.
+
+By the first evaluation (July 29 - July 3), I will hopefully complete module 3
+
+By the second evaulation (July 27 - July 31), I plan to complete adding support for 2 more file formats.
+
+And by Aug 24, I will add support for 2 more, Hopefully I will be able to add support for more file formats as discussed(Module 5)
```
ojwb: I'd really like to see this broken down more by time.  Currently your coding period is split into 2 sections of 3 weeks and 8 weeks, but nothing is mapped out in smaller units of time.  That means that most of the time during the coding period we can't tell how on track we are easily, which is far from ideal.

Also, how have you come up with these lengths of time?  My experience is that people are bad at estimating time for large tasks but can do a reasonable job of estimating small tasks, so breaking up a large task into smaller subtasks until you reach a level of detail you are more familiar with is a good way to get better time estimates.  It also forces you to think more about the details, and the details are usually what actually takes a lot of the time.

proposal.rst
```diff
@@ -260,18 +351,53 @@ Project Timeline
...
+  * Get to know the community and interact with its members
+  * Understand Xapian Code Base, coding conventions and all relevant classes
+  * Understand how the code I am going to add will be structured and what kind of tests i will have to write
+  * Understand which file formats were chosen for previous year's project and why and understand how they were integrated
+  * Discuss with community, other than zip, which file formats are important and are needed to be added to omega.
+
+
+**Coding Weeks 1-3(June 1 - June 20) Module 3**
+
+  * Research about different zip file reading libraries- eg libarchive,libzip ( Compare them and discuss with the community which is the most appropriate)
```
ojwb: I'd encourage you to try to get the bulk of such research done before the coding period starts.

And things involving community discussion will tend to take a lot of elapsed time, so better to start them sooner.

proposal.rst
Exter-dg: Hi @ojwb , apologies for the uninformative title. I’ve edited it now with the name of the project.

proposal.rst
Exter-dg: Thanks for pointing it out. I’ve corrected this mistake in the improved timeline.

proposal.rst
Exter-dg: Thanks for this detailed comment. I tried to use a top-down approach to build a timeline, but due to the uncertainty over the details in the modules, I ended up making two large blocks instead of a more granular level of detail. I’ve updated the timeline to improve this. Please let me know Phase 1 is okay, and if there are other changes I might make to improve its robustness. But I’m having trouble with Phases 2 and 3 (in the updated timeline) because it is unclear which libraries we will add and also because whichever libraries we decide will mostly follow a similar process as Phase 1. Except that it would now take less time for each new library since I would be more familiar with it. So I’m confused about how I should present this. If you could provide some guidance on how I can add more details, that would be very helpful.

proposal.rst
Exter-dg: I guess I underestimated the time it would take to discuss which zip file reading libraries are useful for the community and I’m grateful for your insight that it could take longer. I’m now making this the twin primary focus of the Community Bonding Period along with a project-specific first PR. And I’ve updated the timeline to reflect some of this.

ojwb: It was really a general point about getting these sort of decisions done early, and about community discussion - in the past I've seen students surprised by how long it takes in elapsed time to consult the wider community on things.

In this particular case I'd probably suggest to go with libarchive, since it's mature, widely used, I've used it myself in the past with success, and it supports other archive formats beyond zip files (and being able to index files inside various archive formats would be useful for some people).

proposal.rst
ojwb: I think you're along the right lines there - assuming the libraries don't have additional challenges, you're likely to be a bit quicker for each subsequent one (but plateauing eventually).

I don't think you can do better unless you identify some candidate libraries now and try to assess if they are likely to present additional challenges.

proposal.rst
```diff
@@ -260,18 +347,60 @@ Project Timeline
...
 .. any university classes or exams, vacations, etc), make sure you include them
 .. in your project timeline.
 
-FILLME
+
+
+**Community Bonding Period: (May 4 - June 1) Module 1,2**
+
+  * Get to know the community and interact with its members over the entire bonding period.
+  * Discuss and find a minor issue directly related to the project that needs to be fixed. (2 days)
```
jaylett, line 382: With what you know of the work Bruno did, and the writeup of that project, is there something specific you think needs addressing before you start looking at new file formats?

proposal.rst
```diff
@@ -260,18 +347,60 @@ Project Timeline
...
+  * Identify a solution to the issue and discuss with members. (2 days)
+  * Write code to fix the issue, write relevant tests and documentation as necessary. Familiarize with the development workflow. (3 days)
```
jaylett, line 384: You've already worked on a small PR. What else do you think you need to learn about the workflow? It's helpful to be specific, both because that will help us support you but also because it will help you think about the work you'll need to do.

proposal.rst
```diff
@@ -316,4 +445,4 @@ you plan to use, please give details.**
 .. to clearly identify that code (and keep existing licensing and copyright
 .. details intact), and to check with the mentors that it is OK to use.
 
-FILLME
+I would have to use external libraries. These libraries will belong to free software and be under licenses compatible with MIT/X license and GNU GPL version 2 and later.
```
jaylett: Are all of them compatible with both GPLv2+ and MIT/X?

jaylett (commented):
As olly says, the main thing is you need to break down the timeline more. One suggestion that may help to do this: for the one module (libarchive) you've identified, break that one down (into sections of *at most* one week) and basically say "this is a template for adding support for a number of file formats using one library". Then you can simply identify two or three other libraries that you think are useful. (And it's fine if after later discussion you change which ones they'll be.) That will at least give you a clear timeline to work against, and enable you to think in detail about what steps you'll have to go through to add new formats.

ojwb:

Exter-dg:
> Note you'll need to submit a "final" PDF version of this at https://summerofcode.withgoogle.com/ to be eligible for GSoC (and I don't seem to be able to see one currently, though possibly I'm failing to match github user name and real name).
> 
> You should be able to generate a suitable pdf from the .rst version using `make pdf`.
> 
> You can update the "final" version as many times as you like up to the deadline, so it's wise to submit early and not run the risk of internet problems preventing you submitting in time.

Thanks for the tip! I have uploaded a pdf form of the current version on GSoC's site.

proposal.rst
Exter-dg: > I don't think you can do better unless you identify some candidate libraries now and try to assess if they are likely to present additional challenges.


Thank you for pointing this out, I have tried and identified few file formats which can be worked upon. I have now updated phases 2 and 3 in the timeline with rational with why I am proposing these libraries (these are subject to change), please let me know if this makes sense or should I change something.



> As olly says, the main thing is you need to break down the timeline more. One suggestion that may help to do this: for the one module (libarchive) you've identified, break that one down (into sections of _at most_ one week) and basically say "this is a template for adding support for a number of file formats using one library". Then you can simply identify two or three other libraries that you think are useful. (And it's fine if after later discussion you change which ones they'll be.) That will at least give you a clear timeline to work against, and enable you to think in detail about what steps you'll have to go through to add new formats.

Thanks for the detailed comment. I have tried and broken down Phase 1 into sections and have tried to elaborate these individual sections. If you have any suggestions about the details that I have listed in Phase 1, please let me know, I will work on them.

proposal.rst
Exter-dg: > With what you know of the work Bruno did, and the writeup of that project, is there something specific you think needs addressing before you start looking at new file formats?

With what I know of the work he did, I could not find info about comparing indexing speeds/methods to report indexing speed as compared to external filters.
Maybe we can look that as a small PR in the bonding phase or the testing phase.
Does this make sense of am I missing something obvious?

proposal.rst
Exter-dg: > You've already worked on a small PR. What else do you think you need to learn about the workflow? It's helpful to be specific, both because that will help us support you but also because it will help you think about the work you'll need to do.

Although I have worked on a small PR to get familiar with the workflow, this was tangentially related to the project. I read [https://xapian-developer-guide.readthedocs.io/en/latest/tests/index.html](url) , I do have an idea about how to add libraries but not specifically for testing. So I will require getting familiar with it.

proposal.rst
Exter-dg: No, the libraries that I have proposed belong to the Mozilla Public License (MPL) version 2.0. These are subject to change after discussion with the community.
I have updated the PR.

ojwb:
Thanks for you proposal.

PR #22, opened 2020-03-30
Selin Kurt GSoC 2020-Implement Bindings for Go

sselinkurt:


proposal.rst
```diff
@@ -105,48 +104,46 @@ Background Information
...
 systems for indexed text search.**
 
-FILLME
+	I didn't have enough time to contribute because I just met with Xapian. I will make good use of the time up to the student selection process and contribute to Xapian.
 
 **Tell us about any previous experience with Free Software and Open Source
 other than Xapian.**
 
-FILLME
+	I joined Hacktoberfest.
```
ojwb: And?

What did you do?  What did you learn?  What was good and what was bad about it?

proposal.rst
```diff
@@ -200,27 +195,27 @@ Project Details
...
 FILLME
 
 **How useful will your results be when not everything works out exactly as
 planned?**
 
-FILLME
+	I think my results will be useful because I don't lose motivation in a stress situation.
 
 Project Timeline
 ----------------
```
ojwb, line 223: You've not filled in a timeline for your project at all.

The timeline is a key part of the proposal, as it shows what you're planning to do, and that you've thought about how much you can achieve in the time available.  It also provides a way for us and for you to see how on track the project is as it progresses, which is important for motivation, and for knowing if we need to re-plan if things take longer than expected (or to know that you'll need to find more to do if things take less time than expected).

proposal.rst
```diff
@@ -200,27 +195,27 @@ Project Details
...
-investigation.**
+**Please note any uncertainties or aspects which depend on further research or investigation.**
 
 FILLME
 
 **How useful will your results be when not everything works out exactly as
 planned?**
 
-FILLME
+	I think my results will be useful because I don't lose motivation in a stress situation.
```
ojwb: There's no detail at all here of what you're actually planning to do, nor how you intend to go about it.

The "project ideas" from the list are just ideas, which you need to research and flesh out into a proposal.  You need to do a lot more to make a viable proposal than just point at an idea from the list and add some brief information about yourself.

ojwb:
Thanks for you proposal.

